PR #18, opened 2025-03-07
Support GCP internal ingress schema

tedparagon:
### Issues Closed

- PARA-12073

### Brief Summary

Changes GCP ingress to use GCE instead of Nginx to support switching beween internal and external access.

### Detailed Summary

Updated charts to change ingress attributes when deploying to GCP to use the native GCE load balancer. This provides better scalabilty without requiring compute resources on the cluster. It also switches from the LetsEncrypt certificate manager to the GCP managed certificates for the same reason. A public load balancer will be used `ingress_scheme=external` or a private one when `ingress_scheme=internal`.

A single ingress manifest is used in Terraform instead of the per service ones to minimize costs while also mitigating the need to increase GCP quotas for public IPs or URL Maps.

### Changes

- ingress chart changes
- GCP ingress and certificate Terraform changes

### Unrelated Changes

none

### Future Work

Update the Azure charts and Terraform to support the same functionality.

### Steps to Test

Deploy to GCP and ensure site is accessible and all resources created as expected.

### Screenshots

image

image

image

### DIFF
--- a/gcp/workspaces/paragon/helm/ingress.tf
+++ b/gcp/workspaces/paragon/helm/ingress.tf
@@ -1,129 +1,96 @@
-resource "kubernetes_secret" "microservices" {
-  for_each = var.microservices
+locals {
+  unique_domains = distinct([
+    for service in values(var.public_services) :
+    replace(replace(service.public_url, "https://", ""), "http://", "")
+  ])
+}
 
-  metadata {
-    name      = "${each.key}-secret"
-    namespace = kubernetes_namespace.paragon.id
-  }
+resource "google_compute_managed_ssl_certificate" "cert" {
+  name = "${var.workspace}-certificate"
 
-  lifecycle {
-    ignore_changes = [
-      metadata[0].annotations,
-      metadata[0].labels
-    ]
+  managed {
+    domains = local.unique_domains
   }
 }
 
-resource "helm_release" "cert_manager" {
-  name       = "cert-manager"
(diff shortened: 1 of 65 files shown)


gcp/workspaces/paragon/helm/ingress.tf
```diff
@@ -1,129 +1,96 @@
...
+          }
+        }
+      ]
+    }
+  })
 
   depends_on = [
-    helm_release.cert_manager,
-    time_sleep.wait
+    helm_release.paragon_on_prem,
```
ian-paragon, line 92: Might not be relevant to this PR, but are we keeping it as on_prem? Or should we rename to enterprise?

prepare.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # version of charts, must be semver and doesn't have to match Paragon appVersion
-version="2025.2.20"
+version="2025.03.05"
```
ian-paragon, line 4: Nit pick, should not have a zero to the left

ian-paragon (approved):

gcp/workspaces/paragon/helm/ingress.tf
tedparagon: I don't know that it really matters here but definitely out of scope for this PR.

ishmaelthedestroyer (approved):

PR #22, opened 2025-04-21
feat(infra): added per service environment variable and secret handling

tedparagon:
### Issues Closed

- PARA-13216

### Brief Summary

Adds support for per-service environment variables and secrets. It also supports separating env vars from secrets.

### Detailed Summary

The way environment variable support was initially implemented it put all variables into secrets and didn't support setting values only for specific services. 

So this would work:
``` 
global:
  env:
    WORKER_SHARED_SERVICE_MIN_INSTANCES: 2
    WORKER_SHARED_SERVICE_MAX_INSTANCES: 4
```

But this wouldn’t work:
```
worker-workflows:
  env:
    WORKER_SHARED_SERVICE_MIN_INSTANCES: 5
    WORKER_SHARED_SERVICE_MAX_INSTANCES: 10
```

All of those values would also be treated as secrets which made diagnosing issues and manually updating configs more complex than necessary. This change allows separating the chart `envKeys` list so sensitive values like passwords or access tokens can be declared in a `secretKeys` list instead. Those values will be stored in a secret while values for `envKeys` will be set directly on the pod in clear text. 

It also supports setting the env and/or secret values at the service level so the examples above would both work. 

**NOTE: only a couple charts' `values.yaml` files have had the `envKeys` list split out to use `secretKeys`. I wanted to get validation on the process before updating the rest of the charts since it will be a significant effort.**

### Changes

- update deployments and stateful sets to pull env and secret values from correct place
- update Terraform to set the variables correctly
- add per-service secrets with any overridden secret values
- separate `envKeys` lists on the charts into `envKeys` and `secretKeys` based on sensitivity of the data
- applied secret handling to `paragon-logging` and `paragon-monitoring` which weren't using them before

### Unrelated Changes

- disabled `cache-replay` by default in AWS

### Future Work

- update the lists of `envKeys` and `secretKeys` for each chart based on new config work

### Steps to Test

- deploy service specific env vars and secrets and ensure they get applied as shown below

### Deployment Notes

- manually overriding per-service secret values with `kubectl` will be difficult so the documentation will have to be updated

### Screenshots

`values.yaml`
image

deployed pod description
image


ian-paragon (approved):

loren-paragon (approved):
LGTM

PR #23, opened 2025-04-23
Add support for local YAML-based feature flags

ian-paragon:
Introduced the ability to use locally defined feature flags via YAML instead of the default Git-based configuration. This includes adding support for optional feature flag variables, a `ConfigMap` resource, and corresponding volume mounts in Helm configurations across AWS, Azure, and GCP deployments

### DIFF
--- a/aws/workspaces/paragon/variables.tf
+++ b/aws/workspaces/paragon/variables.tf
@@ -74,6 +74,12 @@ variable "excluded_microservices" {
   default     = []
 }
 
+variable "feature_flags" {
+  description = "Optional base64 encoded feature flags YAML content."
+  type        = string
+  default     = null
+}
+
 variable "ingress_scheme" {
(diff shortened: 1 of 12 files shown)


aws/workspaces/paragon/variables.tf
```diff
@@ -74,6 +74,12 @@ variable "excluded_microservices" {
   default     = []
 }
 
+variable "feature_flags" {
+  description = "Optional base64 encoded feature flags YAML content."
```
tedparagon, line 78: The other repo uses a base64 string for the value since it generates the tfvars file with typescript. We don't do that here. So it would be better to load the file directly. So this variable should just be the path to the `features.yml` file to load.

tedparagon (changes requested):
Need to switch the base64 var to be a file path instead.

tedparagon (approved):

PR #24, opened 2025-04-30
Fix Azure private ingress and chart secrets

tedparagon:
### Issues Closed

- PARA-13069
- PARA-13216
- PARA-13260
- PARA-13612

### Brief Summary

Chart and Terraform changes to improve support for secrets and internal Azure ingress

### Detailed Summary

Updated secrets for OpenObserve and Fluent-Bit so sensitive values are not displayed in pod descriptions.

Cleaned up Terraform `variables.tf` files to exclude `minio` deployments when not used and to make the files more consistent across cloud providers.

Updated Terraform and Helm Charts to support "internal" ingress schemes when using Azure.

Audited chart `values.yaml` files to ensure `envKeys` and `secretKeys` are correctly set including:
- added missing `FEATURE_FLAG_PLATFORM_ENABLED` envKey
- removed legacy `LOG_HERCULES_MONITOR` envKey
- moved some sensitive keys from envKeys to secretKeys

Added history timestamps and additional aliases to bastion startup script. This will make bastion usage clearer and easier.

### Changes

- updated chart `values.yaml` and `ingress.yaml`
- updated terraform variable assignments
- updated Azure terraform internal ingress logic

### Unrelated Changes

- updated bastion startup scripts
- added GitHub PR template file

### Future Work

none

### Steps to Test

- deploy new changes to Azure both with internal and internal-facing ingress schemes
- verify ingresses and load balancers created with correct configs and visibility
- verify that all pods start cleanly without restarts
- verify that previously missing envKeys are now present

### Screenshots

image

image

image


aws/workspaces/infra/templates/bastion/bastion-startup.tpl.sh
```diff
@@ -141,8 +145,29 @@ done
 $(eksctl get iamidentitymapping --cluster ${cluster_name} --arn arn:aws:iam::${aws_account_id}:role/${bastion_role} || eksctl create iamidentitymapping --cluster ${cluster_name} --arn arn:aws:iam::${aws_account_id}:role/${bastion_role} --group system:masters --username ${bastion_role})
 kubectl config set-context --current --namespace=paragon
 
+# Create and configure aliases file
+writeLog "configuring aliases for ubuntu"
+cat > /home/ubuntu/.bash_aliases << 'EOF'
+# Kubernetes aliases
+alias k=kubectl
```
loren-paragon, line 152: so much 😻

aws/workspaces/infra/templates/bastion/bastion-startup.tpl.sh
```diff
@@ -141,8 +145,29 @@ done
...
+alias kg="kubectl get"
+alias kd="kubectl describe"
+alias kl="kubectl logs"
```
loren-paragon, line 155: perhaps add this?

```
kls() {
  local name=$1
  if [ -z "$name" ]; then
    echo "Usage: kls <service-name>"
    return 1
  fi
  kubectl logs -l app.kubernetes.io/name=$name --all-containers=true --prefix=true "$@"
}

```

loren-paragon (approved):
LGTM

aws/workspaces/infra/templates/bastion/bastion-startup.tpl.sh
```diff
@@ -141,8 +145,29 @@ done
...
+alias kx="kubectl exec -it"
+alias ksec="kubectl get secret -n paragon paragon-secrets -o jsonpath='{.data}' | jq -r 'to_entries[] | \"\(.key): \(.value | @base64d)\"' | sort"
+alias kw="watch kubectl get pods"
+alias kwf="watch -- 'kubectl get pods | grep -v fluent'"
+
+# Common aliases
+alias ll="ls -hal"
```
richard-paragon, line 172: hides `.` and `..`
```suggestion
alias ll="ls -hAl"
```

aws/workspaces/infra/templates/bastion/bastion-startup.tpl.sh
richard-paragon: Not sure if this is environment-specific, but I found that `kubectl logs -l app.kubernetes.io/name=$name` had an implicit `--tail` length of 10.  If you see the same behavior, then it might be better to use something like this:
```
kubectl logs  --all-pods --all-containers --prefix deploy/$name "$@"
```

richard-paragon (approved):

aws/workspaces/infra/templates/bastion/bastion-startup.tpl.sh
tedparagon: I'm going to keep this one as-is since the tail can be overridden explicitly and the `deploy/$name` would prevent this from working in statefulset or other cases that might not be deployments.

PR #26, opened 2025-05-09
fix: avoid image pull throttling

tedparagon:
### Issues Closed

- PARA-13663

### Brief Summary

Updated public images to use docker creds to avoid pull limits. Increased logging buffers.

### Detailed Summary

Large deployments (50+ k8s nodes) could run into crash loops trying to pull public images (`fluent-bit` and `flipt`) from Docker Hub anonymously. Adding the `imagePullSecrets` login that is used for the private images avoids this.

`fluent-bit` was also running into cases when logs were potentially lost due to buffer limits. So these were increased.

Increased `zeus` request CPU to match limit since spikes in CPU were causing HPA to autoscale too aggressively.

### Changes

- helm chart changes

### Unrelated Changes

- added `.idea` to `.gitignore`

### Steps to Test

- deploy updated charts and verify all services start cleanly and scale appropriately paying special attention to fluent-bit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ main.tf
 
 # only keep the top level charts
 **/paragon/charts
+
+# ignore IDE files
+.idea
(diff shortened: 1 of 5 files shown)


loren-paragon (approved):
LGTM

ian-paragon (approved):

PR #41, opened 2025-08-14
chore: enabled health-checker monitoring and changed restart time

tedparagon:
### Issues Closed

- PARA-14678

### Brief Summary

Enabled `health-checker` service with public uptime check.

### Detailed Summary

Added support for separate `monitor_path` that Better Stack can hit for status checks that is not the same as the internal k8s health check endpoint.

### Changes

- enabled public `health-checker` check

### Unrelated Changes

- changed restart cron from 5pm EDT to 9am EST since the restarts kept happening during upgrades

### Future Work

- once we are comfortable with the new monitor we can remove the checks on the individual services and also remove public access from microservices that can be private like the `worker-*` ones

### Steps to Test

- deploy and verify Better Stack monitor is created and passing as shown below

### Screenshots

image


loren-paragon (approved):
LGTM

PR #42, opened 2025-08-26
Update azure redis connection strings

tedparagon:
### Issues Closed

- PARA-15154

### Brief Summary

Simplified Azure redis connection string generation.

### Detailed Summary

The connection string logic was overly complicated and therefore difficult to read. This simplifies the logic and also removes the access key from the connection string when not required to better match how it is being used.

### Changes

- terraform output change

### Unrelated Changes

none

### Future Work

none

### Deployment Notes

- The `infra` workspace outputs will have to be regenerated and reapplied to the `paragon` workspace so these changes are used. This is specific to Azure.


franklingarcia-paragon (approved):
LGTM 🚀

PR #49, opened 2025-09-09
fix/PARA-14008/too-many-restarts-alert

franklingarcia-paragon:
### Issues Closed

- Close [PARA-14008](https://useparagon.atlassian.net/browse/PARA-14008)

### Brief Summary

fix prometheus pod restart alert using increase function

### Detailed Summary

The current pod restart alert was using a raw counter metric (kube_pod_container_status_restarts_total) which stays at a fixed number and keeps alerting continuously. This fix replaces the raw counter with the increase() function to detect actual changes in restart counts over a 1-hour window, preventing false positives while still catching real restart issues.

### Changes

- Added PrometheusRule configuration to prometheus subchart values.yaml
- Created prometheusrule.yaml template in prometheus subchart
- Updated alert expression to use increase(kube_pod_container_status_restarts_total[1h]) > 0
- Added proper alert annotations and labels

### QA Notes

The alert will now only trigger when there are actual restart increases, not on persistent counter values

### Deployment Notes

No additional config updates needed - this is a chart-only change

[PARA-14008]: https://useparagon.atlassian.net/browse/PARA-14008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/paragon-monitoring/charts/prometheus/templates/prometheusrule.yaml
+++ b/charts/paragon-monitoring/charts/prometheus/templates/prometheusrule.yaml
@@ -1,0 +1,16 @@
+{{- if and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) .Values.prometheusRule.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ include "prometheus.fullname" . }}-alerts
+  namespace: {{ .Release.Namespace }}
+  labels:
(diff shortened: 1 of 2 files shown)


tedparagon:
The alerts for this originate in Grafana. So the rule change has to happen there. See https://github.com/useparagon/paragon/blob/main/monitors/grafana/templates/alerting/k8-pod-list.yaml#L19

PR #142, opened 2026-04-09
feat(charts): healthz for worker-workflows

franklingarcia-paragon:
### Issues Closed

- [PARA-19760](https://useparagon.atlassian.net/browse/PARA-19760)

### Brief Summary

configurable k8s probe failure thresholds; worker-workflows stricter liveness tolerance

### Detailed Summary

We align with Kubernetes guidance and staff review: keep the same /healthz endpoint for readiness and liveness (full Terminus / dependency checks), and avoid ?isLite=true so we do not mark workers ready or keep them on the load balancer while Redis or other deps are unhealthy. The shared deployment.standard template now optionally emits failureThreshold for liveness and/or readiness only when a subchart sets probes.liveness.failureThreshold or probes.readiness.failureThreshold in its values.yaml, so all other services keep the previous rendered manifests and implicit API defaults (including default failure count). worker-workflows opts in with probes.liveness.failureThreshold: 6 so readiness (implicit threshold 3) tends to drop the pod from the Service before liveness triggers a restart on sustained failures.

### Changes

- charts/paragon-templates/templates/_deployment.yaml: optional probes.liveness.failureThreshold / probes.readiness.failureThreshold from per-chart values.yaml; omit from rendered YAML when unset (defaults unchanged for everyone else).
- charts/paragon-onprem/charts/worker-workflows/values.yaml: ingress.healthcheck_path: /healthz; add probes.liveness.failureThreshold: 6.

### Unrelated Changes

- None

### Future Work

- Optional: add probes.readiness.failureThreshold (or other probe timings) per service in values.yaml if product wants different readiness vs liveness sensitivity.
- Optional: align Deployments with StatefulSets (readycheck_path vs healthcheck_path) for path splits

### Steps to Test

1. Render or deploy charts; for worker-workflows, kubectl get deploy worker-workflows -o yaml (or describe the pod) and confirm:

- readiness and liveness httpGet.path are /healthz.
- liveness includes failureThreshold: 6; readiness has no explicit failureThreshold unless you set it (then expect API default 3). 

2. Pick any other microservice/worker chart: confirm rendered probes do not include failureThreshold lines unless that chart sets probes.*.

### QA Notes

- Full /healthz still returns 503 while the service is terminating (unchanged).
- Behavior for charts without probes in values is intended to match pre-change defaults (same timings; implicit failure thresholds).

### Deployment Notes

- Helm chart-only change after prepare.sh / normal release pipeline; no new tfvars keys for this work.
- If a customer overrides worker-workflows values (e.g. healthcheck_path or probes), those overrides still win.

### Screenshots

[PARA-19760]: https://useparagon.atlassian.net/browse/PARA-19760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/paragon-onprem/charts/worker-workflows/values.yaml
+++ b/charts/paragon-onprem/charts/worker-workflows/values.yaml
@@ -79,6 +79,10 @@ ingress:
   host: '' # When empty, will use chart name + domain
   healthcheck_path: /healthz
 
+probes:
+  liveness:
+    failureThreshold: 6
+
 nodeSelector: {}
 
 tolerations: []
--- a/charts/paragon-templates/templates/_deployment.yaml
+++ b/charts/paragon-templates/templates/_deployment.yaml
@@ -61,6 +61,11 @@ spec:
           {{- $serviceData := fromJson (include "service.inputs" $root) }}
           {{- $category := $serviceData.category | default "" }}
           {{- if or (eq $category "microservice") (eq $category "worker") }}
+          {{- $probes := $root.Values.probes | default dict }}
+          {{- $readinessProbeCfg := index $probes "readiness" | default dict }}
+          {{- $livenessProbeCfg := index $probes "liveness" | default dict }}
+          {{- $readinessFailureThreshold := int (default 3 (index $readinessProbeCfg "failureThreshold")) }}
+          {{- $livenessFailureThreshold := int (default 3 (index $livenessProbeCfg "failureThreshold")) }}
           lifecycle:
             # need to sleep for `deregistration_delay` + 5s to prevent downtime on deployments
             # https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1719#issuecomment-743452334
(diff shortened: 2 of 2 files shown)


tedparagon (approved):

charts/paragon-onprem/charts/worker-workflows/values.yaml
```diff
@@ -77,7 +77,7 @@ ingress:
   enabled: true
   className: ''
   host: '' # When empty, will use chart name + domain
-  healthcheck_path: /healthz
+  healthcheck_path: "/healthz?isLite=true"
```
richard-paragon: I suggest removing this from the PR so that it only contains the hoop changes.

richard-paragon (changes requested):

charts/paragon-onprem/charts/worker-workflows/values.yaml
richard-paragon: I also recommend closing PARA-19760 as "won't do".

What is the right Jira ticket for this PR? Most of the changes are **not** related to PARA-19760.

richard-paragon (approved):

PR #146, opened 2026-05-01
PARA-18448: helm chart rightsizing

tedparagon:
### Issues Closed

- PARA-18448

### Brief Summary

rightsizing default cpu and memory across helm charts

### Detailed Summary

This PR tunes default Kubernetes resource requests and limits across the on-prem application charts, the monitoring stack, and the logging stack so cluster scheduling and cost better match typical steady-state usage. Pheme also picks up explicit worker shared autoscaling env defaults (after removing duplicate default config keys), the umbrella chart disables the health-checker subchart by default, and the prepared chart semver in `prepare.sh` is bumped to `2026.05.01`.

### Changes

- Lower and align default CPU and memory requests/limits for Paragon microservices and observability components (Prometheus, Grafana, exporters, Fluent Bit, OpenObserve, etc.)
- Remove redundant default keys from the Pheme values file and document optional HPA-style comments where autoscaling is disabled
- Turn off the health-checker subchart in the umbrella `paragon-onprem` values so new installs do not deploy it unless opted in
- Bump the chart bundle version string used by `prepare.sh`

### Unrelated Changes

N/A

### Future Work

- Revisit limits after production metrics from rightsized clusters; re-enable or tune health-checker per customer needs
- PARA-20841 - disable pgadmin and redis-insight be default

### Steps to Test

1. Run `./prepare.sh -p <aws|azure|gcp> -t <tag>` and confirm charts copy and version substitution succeeds
2. `helm template` or deploy to a non-prod cluster and verify pods schedule with the new requests and no immediate OOMKilled restarts under light load
3. Confirm intended subcharts (including health-checker when needed) can still be enabled via values overrides

### QA Notes

Health-checker is now disabled by default at the umbrella level; enable it in environment-specific values if your test matrix still requires that workload.

### Deployment Notes

Customers overriding resources in their own values files are unchanged; anyone relying on previous generous defaults should validate autoscaling and limits after upgrade.

### Screenshots

N/A


franklingarcia-paragon (approved):

charts/paragon-logging/charts/fluent-bit/values.yaml
```diff
@@ -199,10 +199,9 @@ readinessProbe:
 
 resources:
   limits:
-    cpu: 0.5
-    memory: 256Mi
+    memory: 100Mi
   requests:
-    cpu: 0.2
+    cpu: 100m
```
richard-paragon, line 204: If `openobserve` is down or slow at responding to API requests (i.e. fluent-bit pushing logs), then `fluent-bit` keeps those logs in memory (because it cannot flush them). This results in fluent-bit OOM when applications are sending a large amount of log data.

You may have mentioned this, but are these numbers based on **average** load or **max** load or some quantile (e.g. P90)?  It's possible that a system is fine for 80% of the time at `50Mi` memory, but needs `300Mi` 20% of the time. The average is `100Mi`, but the system will be crashing 20% of the time.

charts/paragon-monitoring/charts/pgadmin/values.yaml
```diff
@@ -61,22 +61,18 @@ ingress:
...
-  # choice for the user. This also increases chances charts run on environments with little
-  # resources, such as Minikube. If you do want to specify resources, uncomment the following
-  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
   limits:
-    memory: 512Mi
+    memory: 300Mi
   requests:
     cpu: 100m
-    memory: 128Mi
+    memory: 300Mi
```
richard-paragon, line 68: I wonder if this is necessary, given that we're going to be dropping `pgadmin` use by on-call etc.  

Perhaps we could spin this pod up only _on demand_, when hoop isn't sufficient.  

I personally never use `pgadmin`. It's much easier for me to connect to the DBs with `psql` (using a script that sets up port forwarding via the bastion). I also did some testing on spinning up a temporary `psql` **pod** inside a cluster, and that worked really well too. But that's probably less ideal than just connecting with locally-running psql.

Port forwarding also allows us to connect easily with a **locally** running `pgadmin`, so paying for this pod to be running in every bastion seems really unnecessary.

charts/paragon-monitoring/charts/redis-insight/values.yaml
```diff
@@ -61,22 +61,18 @@ ingress:
...
-  # choice for the user. This also increases chances charts run on environments with little
-  # resources, such as Minikube. If you do want to specify resources, uncomment the following
-  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
   limits:
-    memory: 512Mi
+    memory: 200Mi
   requests:
     cpu: 100m
-    memory: 128Mi
+    memory: 200Mi
```
richard-paragon, line 68: I do use `redis-insight`, but I still wonder if this is another case where we can avoid running a pod in the bastions, and just develop better tooling for connecting with a locally-running `redis-insight` or `redis-cli`, or just providing the `kubectl run` command to spin up an on-demand, temporary pod.  (btw secrets for these temporary pods can still be pulled from `paragon-secrets` in the cluster).

charts/paragon-onprem/charts/cache-replay/values.yaml
```diff
@@ -16,18 +16,17 @@ service:
...
 resources:
   limits:
-    cpu: 1
-    memory: 2048Mi
+    memory: 400Mi
   requests:
-    cpu: 0.5
-    memory: 2048Mi
+    cpu: 100m
+    memory: 400Mi
```
richard-paragon: I'm not sure historic values for resource usage are accurate for this service, since we don't run it much and its usage probably depends on how **long** it is intended to be running, as well as the customer-specific load, because it has to store all requests and data in memory for the duration of the outage/downtime, right?

charts/paragon-onprem/charts/zeus/values.yaml
```diff
@@ -16,18 +16,17 @@ service:
...
 resources:
   limits:
-    cpu: 0.5
-    memory: 1024Mi
+    memory: 600Mi
   requests:
-    cpu: 0.5
-    memory: 1024Mi
+    cpu: 200m
+    memory: 600Mi
```
richard-paragon, line 22: `zeus` is so busy, working as the API gateway for all the services. So this reduction is a bit concerning to me. (But maybe it doesn't work as hard in enterprise environments as it does in Cloud)

richard-paragon (changes requested):
No real blocking feedback or critical changes requested. 
Just marking as `Request changes` to avoid it getting auto-merged.
I'm happy to approve as-is once comments have been read/considered.

charts/paragon-logging/charts/fluent-bit/values.yaml
tedparagon: The recommended values are based on the P95 with an additional safety factor (1.15). https://github.com/useparagon-internal/dev-utils/pull/6/changes#diff-6e43e3d032f8e970aef715b5b406db7afd7344bc7dafb886ee997034e2f49f50R61

In the cases where a customer's max usage exceeds that then override values were recommended also. This will handle those outliers with having to oversize the defaults for the P95 cases.

charts/paragon-monitoring/charts/pgadmin/values.yaml
tedparagon: Agreed but I'll create a separate ticket for that and add to future work.

charts/paragon-monitoring/charts/redis-insight/values.yaml
tedparagon: Agreed.

charts/paragon-onprem/charts/zeus/values.yaml
tedparagon: There were a couple that I was surprised about but ended up sticking with the recommended values. We'll have to monitor all of these as the rollout and adjust as needed.

charts/paragon-onprem/charts/cache-replay/values.yaml
tedparagon: Good catch. I reverted this one. If `cache-replay` is actually in use then we definitely don't want it undersized.

richard-paragon (approved):
Looks great. Thanks!

PR #149, opened 2026-05-15
feat/PARA-20977/-arden-helm-deployment-conf-for-azure-managed-service-failover-events

franklingarcia-paragon:
### Issues Closed

- [PARA-20977](https://useparagon.atlassian.net/browse/PARA-20977)

### Brief Summary

harden k8s probes for azure dependency blips

### Detailed Summary

Azure-managed PostgreSQL and Redis maintenance and failover can cause short `/healthz` failures (dependency checks). Default liveness settings (~30s to kill) were too aggressive and contributed to unnecessary pod restarts during those windows (e.g. SEV-754 / SEV-794 class incidents).

This PR centralizes probe rendering in a shared Helm helper, lets Deployments use separate HTTP paths for liveness vs readiness (and optional startup path), exposes full per-probe tuning via `values`, and sets a higher default `probes.liveness.failureThreshold` on Postgres+Redis-backed `paragon-onprem` services so kubelet allows roughly ~3 minutes of consecutive liveness failures before restart while readiness stays stricter by default.

### Changes

- Add `charts/paragon-templates/templates/_probes.tpl` with `probes.httpGet` helper (paths: `ingress.livecheck_path`, `readycheck_path`, `startupcheck_path` with fallbacks; overrides under `probes.{liveness,readiness,startup}.*`).
- Wire `_deployment.yaml` and `_stateful-set.yaml` to use the helper for all three probes (parity between workload kinds).
- Set `probes.liveness.failureThreshold: 18` in default `values.yaml` for all subcharts that depend on Postgres and Redis per `files/service-inputs.json` (includes `pheme`, `worker-eventlogs`, other workers, API/microservices in that set); align `worker-workflows` from 6 to 18.

### Unrelated Changes

- None intentionally. No application code, Terraform, or `prepare.sh` / `scripts` dependency changes as part of this PR.

### Future Work

- Add a dedicated low-dependency liveness HTTP path in services (e.g. `/live`) and set `ingress.livecheck_path` in charts; team policy excludes using `?isLite` on probes.
- Application-side Postgres/Redis retry, backoff, and connection timeouts (separate tickets / monorepo).
- Maintenance-aware alerting and any Azure-specific defaults only if you introduce a stable `values` signal (e.g. cloud provider) without widening probes on other clouds.

### Steps to Test

- From repo root after `prepare.sh` for your provider (or equivalent copy of `charts/` into the workspace charts path Terraform uses), run `helm dependency build` on `paragon-onprem` if needed, then `helm template` with typical `global.env` / image settings and confirm rendered Deployments/StatefulSets show expected `livenessProbe` / `readinessProbe` / `startupProbe` fields (paths and thresholds).
- Spot-check a service with overrides: set `ingress.livecheck_path`, `probes.liveness.periodSeconds`, etc., and confirm rendered YAML matches.
- Optionally `helm lint` on prepared charts per `AGENTS.md` (expect known parent-chart / semver placeholder noise until versions are substituted).

### QA Notes

- Behavior change: liveness tolerates longer consecutive failures before pod restart (~180s with default `periodSeconds: 10` and `failureThreshold: 18`). Readiness defaults unchanged (still fails out of Service relatively quickly). Risk: a truly wedged process may take longer to be killed by liveness alone.
- `/healthz` still performs dependency checks unless a separate live path is configured; this mitigates churn; it does not fix bad client retry behavior.

### Deployment Notes

- No new secrets or ConfigMaps. Rolling deploy will pick up probe changes on next release. No manual config change required beyond normal Helm/Terraform apply for the Paragon chart bundle.

### Screenshots

None

### DIFF
--- a/charts/paragon-onprem/charts/account/values.yaml
+++ b/charts/paragon-onprem/charts/account/values.yaml
@@ -72,6 +72,10 @@ ingress:
   host: '' # When empty, will use chart name + domain
   healthcheck_path: /healthz
 
+probes:
+  liveness:
+    failureThreshold: 18
+
 nodeSelector: {}
 
 tolerations: []
--- a/charts/paragon-onprem/charts/account/values.yaml
+++ b/charts/paragon-onprem/charts/account/values.yaml
@@ -74,7 +74,7 @@ ingress:
 
 probes:
   liveness:
-    failureThreshold: 18
+    failureThreshold: 12
 
 nodeSelector: {}
 
--- a/charts/paragon-templates/templates/_probes.tpl
+++ b/charts/paragon-templates/templates/_probes.tpl
@@ -1,0 +1,56 @@
+{{/*
+Render a Kubernetes container probe (httpGet) for liveness, readiness, or startup.
+
+Usage:
+  livenessProbe:
+    {{- include "probes.httpGet" (dict "root" $ "type" "liveness") | nindent 12 }}
+  readinessProbe:
+    {{- include "probes.httpGet" (dict "root" $ "type" "readiness") | nindent 12 }}
+  startupProbe:
+    {{- include "probes.httpGet" (dict "root" $ "type" "startup") | nindent 12 }}
+
+Path resolution (in order):
+  liveness  -> .Values.ingress.livecheck_path    | .Values.ingress.healthcheck_path | "/healthz"
+  readiness -> .Values.ingress.readycheck_path   | .Values.ingress.healthcheck_path | "/healthz"
+  startup   -> .Values.ingress.startupcheck_path | .Values.ingress.healthcheck_path | "/healthz"
+
+Per-probe overrides (any subset) via .Values.probes.<type>.<field>:
+  initialDelaySeconds, periodSeconds, timeoutSeconds, failureThreshold, successThreshold
+
+Defaults preserve historical chart behavior:
+  liveness, readiness -> initialDelaySeconds: 10, periodSeconds: 10, timeoutSeconds: 5, failureThreshold: 3
+  startup             ->                          periodSeconds: 10, timeoutSeconds: 5, failureThreshold: 30
+*/}}
+{{- define "probes.httpGet" -}}
+{{- $root := .root -}}
+{{- $type := .type -}}
+{{- $ingress := $root.Values.ingress | default dict -}}
+{{- $defaultPath := $ingress.healthcheck_path | default "/healthz" -}}
+{{- $path := $defaultPath -}}
+{{- if eq $type "liveness" -}}
+{{- $path = $ingress.livecheck_path | default $defaultPath -}}
+{{- else if eq $type "readiness" -}}
+{{- $path = $ingress.readycheck_path | default $defaultPath -}}
+{{- else if eq $type "startup" -}}
+{{- $path = $ingress.startupcheck_path | default $defaultPath -}}
+{{- end -}}
+{{- $cfg := dict -}}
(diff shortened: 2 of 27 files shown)


charts/paragon-onprem/charts/account/values.yaml
```diff
@@ -72,6 +72,10 @@ ingress:
   host: '' # When empty, will use chart name + domain
   healthcheck_path: /healthz
 
+probes:
+  liveness:
+    failureThreshold: 18
```
richard-paragon: This seems a bit too long. While the default (30s) is too short, I think 3 minutes might be overdoing it. What is this 3 minute value based on? Is that the length of time the maintenance window takes?

charts/paragon-templates/templates/_probes.tpl
```diff
@@ -0,0 +1,56 @@
...
+{{- else if not $isStartup }}
+initialDelaySeconds: 10
+{{- end }}
+periodSeconds: {{ if hasKey $cfg "periodSeconds" }}{{ int (index $cfg "periodSeconds") }}{{ else }}10{{ end }}
+timeoutSeconds: {{ if hasKey $cfg "timeoutSeconds" }}{{ int (index $cfg "timeoutSeconds") }}{{ else }}5{{ end }}
+{{- if hasKey $cfg "successThreshold" }}
+successThreshold: {{ int (index $cfg "successThreshold") }}
+{{- end }}
+failureThreshold: {{ if hasKey $cfg "failureThreshold" }}{{ int (index $cfg "failureThreshold") }}{{ else if $isStartup }}30{{ else }}3{{ end }}
+{{- end -}}
```
richard-paragon: Not a blocker, but consider whether this would improve maintenance or readability:

Rather than handling each key separately, is there a way to define the default as an object (dict) and then merge in whatever is in `values.yaml`?

---
Example (using a different shaped object)

```go
{{- /* 1. Define your internal default dictionary */ -}}
{{- $defaults := dict "replicaCount" 1 "resources" (dict "requests" (dict "cpu" "100m" "memory" "128Mi")) -}}

{{- /* 2. Merge user-defined values on top of defaults */ -}}
{{- /* mergeOverwrite makes deep-nested updates  */ -}}
{{- $mergedValues := mergeOverwrite $defaults .Values.myConfig -}}

apiVersion: v1
kind: ConfigMap
metadata:
  name: my-app-config
data:
  config.yaml: |
    replicaCount: {{ $mergedValues.replicaCount }}
    resources:
      requests:
        cpu: {{ $mergedValues.resources.requests.cpu }}
        memory: {{ $mergedValues.resources.requests.memory }}
```

richard-paragon (approved):

anand-paragon (approved):

charts/paragon-onprem/charts/account/values.yaml
franklingarcia-paragon: 3 minutes isn’t tied to a specific maintenance window length. It comes from what we’ve seen in past SEVs and on-call notes: Azure maintenance / DB restarts (SEV-302, 586, 654, 754 and similar) often show liveness failures for a few minutes, and Ted noted these alerts usually self-resolve in ~4 minutes. With the default threshold (3 × 10s ≈ 30s), kubelet restarts pods during that blip and makes things worse.

We had also aligned in an earlier thread to start around failureThreshold: 6 (~60s) and expand carefully; I widened it here because the same pattern showed up across several DB-backed services, not only worker-workflows.

franklingarcia-paragon: I just decrease the value to 12 (2 min).

PR #151, opened 2026-05-22
feat/PARA-20444/openobserve-uds-post-hook

franklingarcia-paragon:
### Issues Closed

- [PARA-20444](https://useparagon.atlassian.net/browse/PARA-20444)

### Brief Summary

openobserve uds post-install/post-upgrade helm hook

### Detailed Summary

Added an idempotent Helm post-install/post-upgrade hook on the `openobserve` subchart so the Paragon logs stream user-defined schema (UDS) is applied automatically after OpenObserve is up on each `paragon-logging` install or upgrade. This reduces OpenObserve RAM instability from high-cardinality fields by enforcing a fixed UDS instead of relying on manual runs of the standalone `o2-apply-uds.ts` script.

The hook is a short-lived Job (Alpine + `curl` + `jq`) that reads schema and script from chart `files/`, uses the same OpenObserve credentials as the StatefulSet, waits for `/healthz`, compares the live schema to the desired JSON, and issues a single `PUT` only when needed.

Fixes applied during implementation: Helm `.Files.Get` paths must use `files/<name>` (empty ConfigMap in staging otherwise); the hook script must be POSIX `sh` (no bash `<<<`); no `chmod` on the read-only ConfigMap mount.

### Changes

- add `files/openobserve-uds-schema.json` and `files/apply-uds.sh` under `charts/paragon-logging/charts/openobserve`
- add `templates/uds-configmap.yaml` and `templates/uds-apply-job.yaml` (post-install/post-upgrade hook)
- add `uds.*` values on the openobserve subchart (`enabled`, image, timeouts, resources)
- document the hook in `helm-development.md`

### Unrelated Changes

None.

### Future Work

- batching the `PUT` payload only if the schema grows large enough to stress the OpenObserve API

### Steps to Test

1. Run `./prepare.sh -p <aws|azure|gcp>` from the enterprise repo on a branch that includes this change.
2. Confirm prepared chart contains non-empty UDS assets:
   - `aws/workspaces/paragon/charts/paragon-logging/charts/openobserve/files/openobserve-uds-schema.json`
   - `.../files/apply-uds.sh`
3. `helm upgrade paragon-logging` in a cluster with OpenObserve running (or fresh install).
4. During upgrade, optionally watch: `kubectl get jobs,pods -n paragon -l app.kubernetes.io/component=uds-apply`
5. Verify hook success via events: `Job completed` (not `BackoffLimitExceeded`).
6. Confirm ConfigMap is populated: `kubectl get cm openobserve-uds -n paragon -o yaml` (schema JSON and script present; no `<<<`).
7. Re-run upgrade: hook should complete again; OpenObserve schema should show ~494 `defined_schema_fields` or logs should say `UDS already applied; skipping PUT`.
8. Optional local check: `hoop connect customer_name-openobserve` then `http://127.0.0.1:8999/web`

### QA Notes

- The hook Job is deleted after success (`hook-delete-policy: before-hook-creation,hook-succeeded`); `kubectl get job openobserve-uds-apply` will often return NotFound even when the hook succeeded. Use namespace events or a second upgrade with `kubectl logs` during the run.
- Hook failure fails the Helm release when `atomic = true` on `paragon_logging`.
- Disable with `openobserve.uds.enabled: false` if needed for a one-off bypass.
- Staging may use `paragon-secrets-openobserve` for O2 credentials; the hook uses the same secret selection logic as the OpenObserve StatefulSet.

### Deployment Notes

- Requires a new enterprise chart build after merge: run `prepare.sh` so provider workspaces get updated charts and chart hash/version.
- Deploy via normal Terraform `helm_release.paragon_logging` / `helm upgrade` path; no new container image in `useparagon/*`, no changes to `enterprise/scripts/`.
- No new tfvars required; `openobserve.uds.enabled` defaults to `true`.

### Screenshots
Before:
staging-o2-no-uds
After:
staging-o2-uds-applied

### DIFF
--- a/charts/paragon-logging/charts/openobserve/files/apply-uds.sh
+++ b/charts/paragon-logging/charts/openobserve/files/apply-uds.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 # Apply OpenObserve user-defined schema for the paragon logs stream (idempotent).
-# PARA-20444 — same semantics as o2-apply-uds.ts (GET, compare, single PUT).
+# PARA-20444 — POSIX sh (Alpine /bin/sh); no bash-isms.
 set -eu
 
 DESIRED_SCHEMA="${DESIRED_SCHEMA:-/uds/openobserve-uds-schema.json}"
@@ -19,7 +19,6 @@ if [ ! -f "$DESIRED_SCHEMA" ]; then
   exit 1
 fi
 
-# Strip trailing slash from O2_HOST
 O2_HOST="${O2_HOST%/}"
 
 auth_curl() {
@@ -40,18 +39,18 @@ done
 echo "Fetching current schema"
 current="$(auth_curl "${O2_HOST}/api/default/streams/paragon/schema?type=logs")"
 
-if jq -e --slurpfile d "$DESIRED_SCHEMA" '
+if echo "$current" | jq -e --slurpfile d "$DESIRED_SCHEMA" '
   (.settings.defined_schema_fields // []) as $cur |
   [$d[0].schema[].name] as $want |
   (($want | length) == ($cur | length)) and
   ($want | all(. as $n | $cur | index($n) != null))
-' <<<"$current" >/dev/null; then
+' >/dev/null; then
   echo "UDS already applied; skipping PUT"
   exit 0
 fi
 
 echo "Building UDS payload"
-payload="$(jq -c --slurpfile d "$DESIRED_SCHEMA" '
+payload="$(echo "$current" | jq -c --slurpfile d "$DESIRED_SCHEMA" '
   . as $current |
   $d[0].schema as $desired |
   ($desired | map(.name)) as $add_names |
@@ -67,7 +66,7 @@ payload="$(jq -c --slurpfile d "$DESIRED_SCHEMA" '
       remove: $remove_fields
     }
   }
-' <<<"$current")"
+')"
 
 echo "Applying UDS schema"
 auth_curl \
@@ -78,6 +77,6 @@ auth_curl \
 
 echo "Verifying updated schema"
 updated="$(auth_curl "${O2_HOST}/api/default/streams/paragon/schema?type=logs")"
-uds_count="$(jq '(.uds_schema // []) | length' <<<"$updated")"
-defined_count="$(jq '(.settings.defined_schema_fields // []) | length' <<<"$updated")"
+uds_count="$(echo "$updated" | jq '(.uds_schema // []) | length')"
+defined_count="$(echo "$updated" | jq '(.settings.defined_schema_fields // []) | length')"
 echo "UDS applied: ${uds_count} uds_schema fields, ${defined_count} defined_schema_fields"
--- a/charts/paragon-logging/charts/openobserve/files/apply-uds.sh
+++ b/charts/paragon-logging/charts/openobserve/files/apply-uds.sh
@@ -1,55 +1,94 @@
 #!/bin/sh
-# Apply OpenObserve user-defined schema for the paragon logs stream (idempotent).
-# PARA-20444 — POSIX sh (Alpine /bin/sh); no bash-isms.
+# OpenObserve UDS apply hook (PARA-20444) — POSIX sh for Alpine /bin/sh.
+#
+# When to run: Helm post-install / post-upgrade on paragon-logging (OpenObserve must be up).
+#
+# Process:
+#   1. Load the desired field list from openobserve-uds-schema.json (chart files/).
+#   2. GET the live "paragon" logs stream schema from OpenObserve.
+#   3. If settings.defined_schema_fields already matches desired names, skip PUT (idempotent).
+#   4. Otherwise build a PUT body:
+#        - add: every desired field name + field definitions
+#        - remove: names/types present on the server but not in the desired list
+#   5. PUT stream settings, then GET again and fail unless counts match the JSON.
+#
+# All progress is written to stdout for hook Job logs (kubectl logs / cluster events).
 set -eu
 
+LOG_PREFIX="[openobserve-uds]"
 DESIRED_SCHEMA="${DESIRED_SCHEMA:-/uds/openobserve-uds-schema.json}"
 O2_HOST="${O2_HOST:-http://openobserve:5080}"
 O2_USER="${O2_USER:-${ZO_ROOT_USER_EMAIL:-}}"
 O2_PASS="${O2_PASS:-${ZO_ROOT_USER_PASSWORD:-}}"
 HEALTH_WAIT_SECONDS="${HEALTH_WAIT_SECONDS:-120}"
 
-if [ -z "$O2_USER" ] || [ -z "$O2_PASS" ]; then
-  echo "error: O2_USER/O2_PASS (or ZO_ROOT_USER_EMAIL/ZO_ROOT_USER_PASSWORD) required" >&2
+log() {
+  echo "${LOG_PREFIX} $*"
+}
+
+die() {
+  echo "${LOG_PREFIX} error: $*" >&2
   exit 1
+}
+
+if [ -z "$O2_USER" ] || [ -z "$O2_PASS" ]; then
+  die "O2_USER/O2_PASS (or ZO_ROOT_USER_EMAIL/ZO_ROOT_USER_PASSWORD) required"
 fi
 
 if [ ! -f "$DESIRED_SCHEMA" ]; then
-  echo "error: desired schema not found: $DESIRED_SCHEMA" >&2
-  exit 1
+  die "desired schema not found: ${DESIRED_SCHEMA}"
 fi
 
 O2_HOST="${O2_HOST%/}"
+
+expected_count="$(jq '.schema | length' "$DESIRED_SCHEMA")"
+log "starting UDS apply for stream paragon (desired fields: ${expected_count})"
+log "openobserve endpoint: ${O2_HOST}"
 
 auth_curl() {
   curl -sf -u "$O2_USER:$O2_PASS" "$@"
 }
 
-echo "Waiting for OpenObserve at ${O2_HOST}/healthz (max ${HEALTH_WAIT_SECONDS}s)..."
+schema_matches_desired() {
+  echo "$1" | jq -e --slurpfile d "$DESIRED_SCHEMA" '
+    (.settings.defined_schema_fields // []) as $cur |
+    [$d[0].schema[].name] as $want |
+    (($want | length) == ($cur | length)) and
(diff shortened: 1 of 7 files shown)


charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -0,0 +1,82 @@
...
+
+DESIRED_SCHEMA="${DESIRED_SCHEMA:-/uds/openobserve-uds-schema.json}"
+O2_HOST="${O2_HOST:-http://openobserve:5080}"
+O2_USER="${O2_USER:-${ZO_ROOT_USER_EMAIL:-}}"
+O2_PASS="${O2_PASS:-${ZO_ROOT_USER_PASSWORD:-}}"
+HEALTH_WAIT_SECONDS="${HEALTH_WAIT_SECONDS:-120}"
+
+if [ -z "$O2_USER" ] || [ -z "$O2_PASS" ]; then
+  echo "error: O2_USER/O2_PASS (or ZO_ROOT_USER_EMAIL/ZO_ROOT_USER_PASSWORD) required" >&2
+  exit 1
```
richard-paragon: Will this show up during the installation upgrade in a way that you/Ted will be able to see it?

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -0,0 +1,82 @@
...
+  -X PUT \
+  -H "Content-Type: application/json" \
+  -d "$payload" \
+  "${O2_HOST}/api/default/streams/paragon/settings"
+
+echo "Verifying updated schema"
+updated="$(auth_curl "${O2_HOST}/api/default/streams/paragon/schema?type=logs")"
+uds_count="$(echo "$updated" | jq '(.uds_schema // []) | length')"
+defined_count="$(echo "$updated" | jq '(.settings.defined_schema_fields // []) | length')"
+echo "UDS applied: ${uds_count} uds_schema fields, ${defined_count} defined_schema_fields"
```
richard-paragon: Should we compare the counts and exit with an error if they are not equal?

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/sh
+# Apply OpenObserve user-defined schema for the paragon logs stream (idempotent).
+# PARA-20444 — POSIX sh (Alpine /bin/sh); no bash-isms.
```
richard-paragon: Why? It seems like this was an arbitrary restriction. Can you just use a `node:22` image?

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -0,0 +1,82 @@
...
+done
+
+echo "Fetching current schema"
+current="$(auth_curl "${O2_HOST}/api/default/streams/paragon/schema?type=logs")"
+
+if echo "$current" | jq -e --slurpfile d "$DESIRED_SCHEMA" '
+  (.settings.defined_schema_fields // []) as $cur |
+  [$d[0].schema[].name] as $want |
+  (($want | length) == ($cur | length)) and
+  ($want | all(. as $n | $cur | index($n) != null))
```
richard-paragon: It might be hard for readers to know what is happening here.

Consider adding a comment at the top of the script that describes the whole process and explains why we need to pull the current fields and find the intersection/complement of the two lists (existing fields and desired fields).

It would also be helpful to add another comment here explaining what `jq` is doing specifically in these lines, and then later in the "Building UDS payload".

richard-paragon (approved):

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: Fair push — not arbitrary, but a tradeoff for this ticket.

I used Alpine + `curl`/`jq` to avoid a heavier runtime image and any extra build/publish step for a small hook. The script is POSIX `sh` because the Job runs `/bin/sh` (ash), not bash — we actually hit that in staging when the first version used `<<<`.

`node:22` is reasonable if we want to align with the manual `o2-apply-uds.ts` and centralize logic long term. I optimized here for “no custom image, minimal surface.” Happy to switch in a follow-up if you’d rather standardize on Node for maintainability.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: Good call — it was too quiet before.

In `62b88f8` the script logs to stdout with a `[openobserve-uds]` prefix: health wait, before/after field counts, payload add/remove sizes, and `done` / `done (no changes)`. That should be visible in `kubectl logs` on the hook pod during the upgrade window (while the pod still exists), plus events like `Job completed`.

If we want easier post-mortem in non-prod, I can add `ttlSecondsAfterFinished` in a follow-up so the pod sticks around a bit longer after success.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: Agreed — added in `62b88f8`:

- A header comment describing the full flow (GET → compare → PUT → verify) and why we diff desired vs existing fields.
- A note above the payload `jq` block explaining add/remove (same semantics as `o2-apply-uds.ts`).

Let me know if you want more inline comments on the idempotency check block as well.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: Yes — implemented in `62b88f8`.

After the PUT we re-GET the schema and exit 1 if:
- `defined_schema_fields` count ≠ desired JSON field count, or
- `uds_schema` count ≠ same expected count, or
- the name set still doesn’t match the desired list.

That way a partial/failed apply fails the hook (and the release when `atomic` is on) instead of only printing counts.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
richard-paragon: I'm not sure what you mean by post-mortem.  What I was asking is if this would throw an error causing the upgrade to fail, so the person doing the install sees that it failed.
Just writing a message to a log probably is not going to be noticed,

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
richard-paragon: When you say "heavier runtime image", what do you mean? How much heavier is node:22-alpine vs alpine? How much longer does it take to spin up?  This is a job that gets executed once a month, right?

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
richard-paragon: You keep mentioning `62b88f8`, but that doesn't appear to be a commit hash prefix. Maybe it hasn't been pushed yet?

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/sh
+# Apply OpenObserve user-defined schema for the paragon logs stream (idempotent).
+# PARA-20444 — POSIX sh (Alpine /bin/sh); no bash-isms.
+set -eu
+
+DESIRED_SCHEMA="${DESIRED_SCHEMA:-/uds/openobserve-uds-schema.json}"
+O2_HOST="${O2_HOST:-http://openobserve:5080}"
+O2_USER="${O2_USER:-${ZO_ROOT_USER_EMAIL:-}}"
+O2_PASS="${O2_PASS:-${ZO_ROOT_USER_PASSWORD:-}}"
```
richard-paragon: nit: You can probably just remove O2_USER, O2_PASS and just use ZO_ROOT_USER_EMAIL/ZO_ROOT_USER_PASSWORD

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/sh
+# Apply OpenObserve user-defined schema for the paragon logs stream (idempotent).
+# PARA-20444 — POSIX sh (Alpine /bin/sh); no bash-isms.
+set -eu
+
+DESIRED_SCHEMA="${DESIRED_SCHEMA:-/uds/openobserve-uds-schema.json}"
```
richard-paragon: nit  DESIRED_SCHEMA --> DESIRED_SCHEMA_FILE

anand-paragon (approved):
this looks good to mer, But can we also consult team leads on the schema before merging this

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: I meant fail the release, not rely on logs. The hook exits non-zero on timeout/PUT/verification failure, and paragon_logging uses atomic = true, so the upgrade should fail visibly.
Pushed 62b88f8 with post-PUT validation; sorry I replied before that was on the branch.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: Maybe “heavier” was vague. Roughly alpine:3.20 ~14MB vs node:22-alpine ~229MB; for a monthly job, Node is reasonable if we want parity with o2-apply-uds.ts.
I kept Alpine+curl+jq for this ticket to avoid a custom image; happy to switch in a follow-up if you want Node standardization.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: You’re right — I referenced it before push. It’s on the branch now (62b88f8 + 6ea7d3c for the nits).
Thanks for flagging.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: Done in 6ea7d3c — script uses ZO_ROOT_USER_EMAIL / ZO_ROOT_USER_PASSWORD directly, same as the StatefulSet.
Thanks.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
franklingarcia-paragon: Renamed in 6ea7d3c (script + Job env).
Thanks.

PR #152, opened 2026-05-27
Feat/para 21339/grafana hoop connection

franklingarcia-paragon:
### Issues Closed

- [PARA-21339](https://useparagon.atlassian.net/browse/PARA-21339)

### Brief Summary

optional hoop tcp connection to private grafana

### Detailed Summary

Adds an opt-in `hoop_grafana_connection` flag across AWS, Azure, and GCP paragon workspaces. When enabled, Terraform provisions a Hoop TCP application connection to `grafana.paragon:4500` so operators can reach the private Grafana UI through Hoop without exposing it publicly (e.g. when Grafana is listed under `private_services`).

Customers rely on private Grafana and need this connection for operator access via Hoop.

Also bumps the GCP `google` provider to `>= 7.19.0` in the infra workspace so `broker_capacity_config` on `google_managed_kafka_cluster` is supported (required for current GMK Terraform resources).

### Changes

- add `hoop_grafana_connection` variable (default `false`) and wire it through paragon `modules.tf` on aws, azure, and gcp
- add conditional `grafana` Hoop TCP connection in `hoop/locals.tf` (host `grafana.paragon`, port `4500`, monitoring tags) on all three clouds
- bump `hashicorp/google` provider constraint to `>= 7.19.0, < 8.0.0` in `gcp/workspaces/infra` and refresh `.terraform.lock.hcl`

### Unrelated Changes

- gcp infra google provider / lockfile update is bundled for GMK `broker_capacity_config` support; it is not specific to the hoop grafana connection but ships in the same branch for terraform apply compatibility on gcp.

### Future Work

- document `hoop_grafana_connection` in customer deployment / tfvars examples if we standardize enabling it per tenant
- consider aligning grafana host/port with values from monitoring chart outputs instead of hardcoded `grafana.paragon:4500` if those differ by environment

### Steps to Test

1. on a branch with hoop enabled, set `hoop_grafana_connection = true` in paragon tfvars (or equivalent secure values) and run `terraform plan` in `*/workspaces/paragon` for aws/azure/gcp — confirm a new `grafana` hoop connection appears in plan
2. with flag `false` (default), confirm plan shows no grafana connection diff
3. after apply, in Hoop UI verify `{prefix}-grafana` TCP connection exists and connect reaches Grafana on port 4500 inside the cluster
4. for gcp infra: `cd gcp/workspaces/infra && terraform init -backend=false && terraform validate` and confirm provider resolves at `>= 7.19.0`

### QA Notes

- feature is **opt-in** (`hoop_grafana_connection` defaults to `false`); no behavior change until explicitly enabled per deployment
- requires hoop agent/workspace already configured; grafana must be reachable at `grafana.paragon:4500` from the hoop agent network context
- not a customer-facing UI change — operator / internal access path only

### Deployment Notes

- other deployments: leave `false` unless private Grafana + Hoop access is needed
- gcp customers using infra workspace need provider lock refresh on apply (`terraform init`); no application chart version bump required for this change alone

### Screenshots

_n/a — terraform / hoop connection configuration only_

[PARA-21339]: https://useparagon.atlassian.net/browse/PARA-21339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aws/workspaces/paragon/hoop/variables.tf
```diff
@@ -57,6 +57,12 @@ variable "hoop_redis_guardrail_rules" {
   default     = ["182f59b2-5d5d-4ab8-978e-94472b3915fc"]
 }
 
+variable "hoop_grafana_connection" {
+  description = "Whether to create a Hoop TCP connection to Grafana (grafana.paragon:4500)."
+  type        = bool
+  default     = false
```
tedparagon, line 63: Any reason to not enable this by default?

tedparagon (approved):
One nonblocking question.

richard-paragon (approved):

aws/workspaces/paragon/hoop/variables.tf
franklingarcia-paragon: We kept it false by default because only a couple of deployments need it today.

Private Grafana behind Hoop isn’t something we run for every customer workspace. As far as we know, # and # are the only ones with that setup, so they’re the ones that should opt in via hoop_grafana_connection = true in their tfvars. For everyone else, turning this on by default would just create an extra Hoop TCP connection that isn’t used.

Happy to flip the default later if private Grafana becomes standard across more customer repos.

PR #178, opened 2026-06-30
Support private container registry and Artifactory image proxy

tedparagon:
### Issues Closed

- PARA-21725

### Brief Summary

private registry overrides for all helm workloads

### Detailed Summary

Enterprise customers often pull Paragon images through a private registry proxy (e.g. Artifactory) instead of public upstreams. This change adds centralized Helm helpers to rewrite image references, merge pull secrets globally, and optionally skip the default `useparagon/` prefix when mirroring. Terraform gains configurable pull-secret creation so customers can pre-provision credentials and reference them from Helm values.

### Changes

- Add `paragon-templates` helpers for image resolution, pull secrets, test/kubectl images, and a shared test-connection template
- Wire deployments, statefulsets, migrations, restart cron, and helm tests through the new helpers
- Add `global.imageRegistry`, `imageRepositoryPrefix`, `imagePullSecrets`, `testImage`, and `kubectlImage` values
- Make registry pull-secret creation optional and configurable via `docker_pull_secret_name` and `create_docker_pull_secret` across AWS, Azure, and GCP
- Document private registry configuration in README

### Unrelated Changes

N/A

### Future Work

N/A

### Steps to Test

1. Run `./prepare.sh -p aws -t <tag>` and `helm template` with sample `global.imageRegistry` / `imageRepositoryPrefix` values; verify rendered manifests point at the expected registry host
2. Grep rendered output for unexpected upstream hosts (`docker.io`, `ghcr.io`, `public.ecr.aws`)
3. Run `terraform validate` in `{aws,azure,gcp}/workspaces/paragon` with `create_docker_pull_secret = false` and confirm the docker login secret resource is omitted

### QA Notes

When `imageRepositoryPrefix` is omitted, `useparagon/` paths are preserved for mirrored layouts that keep that prefix. Setting it to `""` strips `useparagon/` without adding a replacement prefix.

### Deployment Notes

Customers using a private registry should set matching values in `.secure/values.yaml` (`global.imageRegistry`, `global.imagePullSecrets`) and `vars.auto.tfvars` (`docker_registry_server`, optionally `create_docker_pull_secret = false` and `docker_pull_secret_name`).

### Screenshots

N/A

---

> [!NOTE]
> **Medium Risk**
> Wide Helm and Terraform changes affect every workload image reference and pull credentials; misaligned registry host or pull-secret settings could block cluster rollouts, though defaults preserve current Docker Hub behavior.
> 
> **Overview**
> Adds **private registry / Artifactory proxy** support so enterprise deployments can pull all Paragon images through a customer mirror instead of public upstreams.
> 
> **Helm** introduces shared `paragon-templates` helpers (`paragon.image`, `paragon.imagePullSecrets`, `paragon.containerImage`, `test-connection.standard`, etc.) and wires deployments, statefulsets, migrations, jobs, logging/monitoring subcharts, and the restart cron through them. New `global` knobs include `imageRegistry`, optional `imageRepositoryPrefix` (rewrite or strip `useparagon/`), `migrationImageRegistry`, merged `imagePullSecrets`, and overrides for helm-test **busybox** and cron **kubectl** images. Per-chart hardcoded `docker-cfg` pull secrets are removed in favor of global merge.
> 
> **Terraform** (AWS, Azure, GCP paragon workspaces) adds `docker_pull_secret_name` and `create_docker_pull_secret`; the registry login secret is optional and, when created, its name is concatenated into Helm `global.imagePullSecrets`. `docker_registry_server` is documented to align with `global.imageRegistry` host. Root **README** documents the values + tfvars pairing; chart prep version bumps to `2026.06.30`; generated READMEs bump Terraform to `>= 1.9.0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4a50c3c509d000e8a8fd4533c231697d6f994c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/charts/paragon-logging/charts/fluent-bit/templates/_pod.tpl
+++ b/charts/paragon-logging/charts/fluent-bit/templates/_pod.tpl
@@ -3,6 +3,12 @@ serviceAccountName: {{ include "fluent-bit.serviceAccountName" . }}
 {{- with .Values.imagePullSecrets }}
 imagePullSecrets:
   {{- toYaml . | nindent 2 }}
+{{- else }}
+{{- $pullSecrets := include "paragon.imagePullSecrets" (dict "root" .) -}}
+{{- if $pullSecrets }}
+imagePullSecrets:
+  {{- $pullSecrets | nindent 2 }}
+{{- end }}
 {{- end }}
 {{- if .Values.priorityClass.name }}
 priorityClassName: {{ .Values.priorityClass.name }}
@@ -38,7 +44,7 @@ containers:
     securityContext:
       {{- toYaml . | nindent 6 }}
   {{- end }}
-    image: {{ include "fluent-bit.image" (merge .Values.image (dict "tag" (default .Chart.AppVersion .Values.image.tag))) | quote }}
+    image: {{ include "fluent-bit.image" (merge .Values.image (dict "tag" (default .Chart.AppVersion .Values.image.tag) "root" .)) | quote }}
     imagePullPolicy: {{ .Values.image.pullPolicy }}
   {{- if or .Values.env .Values.envWithTpl }}
     env:
@@ -103,7 +109,7 @@ containers:
     {{- end }}
 {{- if .Values.hotReload.enabled }}
   - name: reloader
-    image: {{ include "fluent-bit.image" .Values.hotReload.image }}
+    image: {{ include "fluent-bit.image" (merge .Values.hotReload.image (dict "root" .)) | quote }}
     args:
       - {{ printf "-webhook-url=http://localhost:%s/api/v2/reload" (toString .Values.metricsPort) }}
       - -volume-dir=/watch/config
--- a/charts/paragon-logging/charts/fluent-bit/templates/_pod.tpl
+++ b/charts/paragon-logging/charts/fluent-bit/templates/_pod.tpl
@@ -1,14 +1,9 @@
 {{- define "fluent-bit.pod" -}}
 serviceAccountName: {{ include "fluent-bit.serviceAccountName" . }}
-{{- with .Values.imagePullSecrets }}
-imagePullSecrets:
-  {{- toYaml . | nindent 2 }}
-{{- else }}
 {{- $pullSecrets := include "paragon.imagePullSecrets" (dict "root" .) -}}
 {{- if $pullSecrets }}
 imagePullSecrets:
   {{- $pullSecrets | nindent 2 }}
-{{- end }}
 {{- end }}
 {{- if .Values.priorityClass.name }}
 priorityClassName: {{ .Values.priorityClass.name }}
--- a/charts/paragon-logging/charts/openobserve/templates/statefulset.yaml
+++ b/charts/paragon-logging/charts/openobserve/templates/statefulset.yaml
@@ -21,6 +21,11 @@ spec:
       labels:
         {{- include "openobserve.selectorLabels" . | nindent 8 }}
     spec:
+      {{- $pullSecrets := include "paragon.imagePullSecrets" (dict "root" .) -}}
+      {{- if $pullSecrets }}
+      imagePullSecrets:
+        {{- $pullSecrets | nindent 8 }}
+      {{- end }}
       serviceAccountName: {{ include "openobserve.serviceAccountName" . }}
       securityContext:
         fsGroup: 2000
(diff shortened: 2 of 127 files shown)


charts/paragon-templates/templates/_helpers.tpl
```diff
@@ -0,0 +1,146 @@
...
+{{- $first := index $parts 0 | default $repository -}}
+{{- if include "paragon.isRegistryHost" $first -}}
+{{- $repository -}}
+{{- else if and $root.Values.global (hasKey $root.Values.global "imageRepositoryPrefix") -}}
+{{- if hasPrefix "useparagon/" $repository -}}
+{{- $repository = trimPrefix "useparagon/" $repository -}}
+{{- end -}}
+{{- with $root.Values.global.imageRepositoryPrefix -}}
+{{- $repository = printf "%s/%s" . $repository -}}
+{{- end -}}
```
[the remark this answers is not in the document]
tedparagon: This is by design. All images should be pulled from the repository and that includes third party ones.

franklingarcia-paragon (approved):

charts/paragon-logging/charts/fluent-bit/templates/_pod.tpl
```diff
@@ -103,7 +104,7 @@ containers:
     {{- end }}
 {{- if .Values.hotReload.enabled }}
   - name: reloader
-    image: {{ include "fluent-bit.image" .Values.hotReload.image }}
+    image: {{ include "fluent-bit.image" (merge .Values.hotReload.image (dict "root" .)) | quote }}
```
richard-paragon, line 107: This merge looked a little suspicious at first, because I thought `.Values.hotReload.image` was being merged with `$root`, but now I get that `root: $root` is a property in the merged object.

Still, I think it might be a little cleaner if the image were passed in as a separate property (as shown below), although it would cause more churn in the `fluent-bit.image` template:


```suggestion
    image: {{ include "fluent-bit.image" (dict "root" . "image"  .Values.hotReload.image) | quote }}
```

charts/paragon-logging/charts/openobserve/templates/statefulset.yaml
```diff
@@ -21,6 +21,11 @@ spec:
       labels:
         {{- include "openobserve.selectorLabels" . | nindent 8 }}
     spec:
+      {{- $pullSecrets := include "paragon.imagePullSecrets" (dict "root" .) -}}
+      {{- if $pullSecrets }}
+      imagePullSecrets:
+        {{- $pullSecrets | nindent 8 }}
+      {{- end }}
```
richard-paragon, line 28: Is there a reason this `if` block can't be output by the `paragon.imagePullSecrets` template?
Is it because the indentation varies from one usage to another?

It looks like this block is copied in a lot of places. Maybe there should be two templates: `paragon.imagePullSecrets` and `paragon.imagePullSecretsBlock`.  The former would remain as it is now, and the latter would include the block of code here (setting `$pullSecrets` and conditionally outputting `imagePullSecrets`)

charts/paragon-templates/templates/_helpers.tpl
richard-paragon, line 31: nit: not sure if the `with` is really necessary here


```suggestion
{{- $repository = printf "%s/%s" $root.Values.global.imageRepositoryPrefix $repository -}}
```

richard-paragon (approved):
Looks great. Also nice cleanup of `test-connection` manifests.
Couple minor nit comments, but no blockers.

charts/paragon-logging/charts/fluent-bit/templates/_pod.tpl
tedparagon: Good call on readability — the `merge` pattern is a bit opaque. I kept it as-is here because `fluent-bit.image` expects flat `repository`/`tag`/`digest` fields at the top level, and the main container on line 42 uses the same pattern.

Switching to `(dict "root" . "image" .Values.hotReload.image)` would be cleaner at the call site, but it means updating `fluent-bit.image` (and every call site) to read from `.image.*` instead. Happy to do that as a small follow-up if you'd prefer — didn't want to expand scope on this PR.

charts/paragon-templates/templates/_helpers.tpl
tedparagon: The `with` is intentional here — it skips prefixing when `imageRepositoryPrefix` is an empty string. That's the documented behavior in the PR QA notes: setting it to `""` strips `useparagon/` without adding a replacement prefix.

An unconditional `printf "%s/%s" $root.Values.global.imageRepositoryPrefix $repository` would produce paths like `/account` (leading slash) when the prefix is empty.

charts/paragon-logging/charts/openobserve/templates/statefulset.yaml
tedparagon: Yes — indentation is the main reason. The block is duplicated ~15 times but `nindent` varies by context (2 for fluent-bit pod, 4 for test hooks, 8 for deployments/statefulsets, 12 for restart-cron). A `paragon.imagePullSecretsBlock` helper would need an indent param, e.g. `(dict "root" . "indent" 8)`.

kube-state-metrics deployment also uses a slightly different guard (`if or .Values...` + `indent` vs checking rendered output). Worth consolidating in a follow-up — didn't want to churn every template in this PR.

PR #200, opened 2026-07-24
feat(PARA-21683): Cloud Armor WAF + rate limiting for GCP public ingress

franklingarcia-paragon:
### Issues Closed

- [PARA-21683](https://useparagon.atlassian.net/browse/PARA-21683)

### Brief Summary

_cloud armor waf and rate limiting for gcp public ingress_

### Detailed Summary

External scan-style traffic reaches Paragon public ingress directly and has caused pod restarts. AWS already has a WAF layer (`aws/workspaces/paragon/waf`) driven by object-typed variables; this brings the same capability to GCP.

A new `waf` submodule builds a **Cloud Armor backend security policy** and the `helm` module attaches it to the public backend services through a `BackendConfig`. GKE already uses native GCE Ingress (an external Application Load Balancer), so Cloud Armor attaches to what is already there without introducing a new component — the same shape as AWS, where WAF v2 attaches to the existing ALB.

The variable surface follows the AWS conventions: empty defaults at the root, one map or object per rule family, and validation on every input so misconfiguration fails at plan time instead of at apply.

Everything is gated by `waf_enabled`, which defaults to `false`, so existing deployments see no plan diff.

### Changes

- `gcp/workspaces/paragon/waf/` — new submodule rendering `google_compute_security_policy` from the variables below
- `waf_enabled` — feature flag, `false` by default
- `waf_ip_whitelist` / `waf_ip_blacklist` / `waf_ip_blacklist_deny_status` — allow and deny lists, bare addresses normalized to `/32` or `/128`, chunked 10 CIDRs per rule which is the Cloud Armor per-rule maximum
- `waf_rate_limit_global` and `waf_rate_limit_paths` with their window variables — per-key limits, path prefixes evaluated before the global limit
- `waf_rate_limit_options` — shared behaviour for the generated limits: throttle vs `rate_based_ban`, enforce key, ban thresholds, preview
- `waf_preconfigured_rules` — Google-managed rule sets, the Cloud Armor equivalent of AWS managed rule groups, with sensitivity, opt-in and opt-out signature IDs, and field exclusions
- `waf_custom_rules` — arbitrary CEL rules, each with its own action and optional rate limit
- `waf_advanced_options` — JSON body parsing, log level, user IP headers, Adaptive Protection
- `waf_logs_sample_rate` — request log sampling on the protected backends
- `helm/ingress.tf` — `paragon-waf-backendconfig` manifest; Grafana's existing BackendConfig receives the same `securityPolicy` block alongside its custom health check
- `helm/helm.tf` — public microservices and monitors annotated with the shared BackendConfig
- `helm/helm-managed-sync.tf` — the managed-sync chart publishes its own Ingress for `sync.${domain}`, so its `api-sync` Service is annotated too
- `outputs.tf` — `waf_security_policy_name` and `waf_rule_count`

Generated rules occupy fixed priority bands so adding config never renumbers unrelated rules: allowlist `1000`, denylist `2000`, path limits `3000`, global limit `4000`, preconfigured `5000`, custom `6000`, default allow at `2147483647`. Explicit `priority` values are honoured, with a `precondition` that fails the plan on duplicates and a `check` that warns when the policy exceeds the default quota of 200 rules.

`paragon-waf-backendconfig` is created whether or not WAF is enabled. An empty `securityPolicy.name` is what tells the GKE controller to **detach** a policy, so dropping the manifest when the flag is off would leave a policy attached and block its destroy.

### Unrelated Changes

`gcp/workspaces/paragon/README.md` is generated by `terraform-docs` and was already stale, so regenerating it also picks up rows from earlier merges: the Secret Manager and ESO resources, and the description changes on `docker_*` and `infra_json*`. Only the `waf_*` rows come from this PR.

### Future Work

- Alerting on Cloud Armor denials from the request logs, so a rule that starts blocking legitimate traffic is noticed without reading logs by hand
- `waf_advanced_options.adaptive_protection_enabled` is exposed but requires Cloud Armor Enterprise, so it is untested here
- Disabling the flag relies on the GKE controller detaching the policy before Terraform destroys it. The ordering is correct, but the controller detaches asynchronously, so a destroy that fails with `resourceInUseByAnotherResource` needs apply to be re-run. Documented on `waf_enabled`; worth revisiting if it turns out to be common.

### Steps to Test

There is no GCP environment available to deploy against, so verification was done offline.

1. `terraform validate` and `terraform fmt -check` on `gcp/workspaces/paragon`.
2. `terraform plan` against a scratch harness with a rich config — IP lists spanning multiple chunks, global and per-path limits, three preconfigured rule sets with exclusions and opt-outs, custom CEL rules with and without rate limits — and confirm the rendered rules, priorities, actions and CEL expressions.
3. Empty config produces exactly one rule, the default allow.
4. Duplicate priorities fail the plan with the intended message.
5. Every root variable validation exercised with a bad value, including `""` and `"/"` as `waf_rate_limit_paths` keys.

On a real cluster, the sequence would be: apply with `waf_enabled = false` and confirm no diff beyond the BackendConfig manifest, then enable with `preview = true` on the rule sets, confirm `gcloud compute security-policies describe` matches the intended rules, watch request logs for would-be blocks, then drop preview.

### QA Notes

Nothing to test while `waf_enabled` is `false`, which is the default for every existing deployment.

When enabling on a customer, start with `preview = true` on `waf_preconfigured_rules` so rules log instead of block. Enabling the SQLi or XSS sets against Paragon's JSON APIs also wants `waf_advanced_options.json_parsing = "STANDARD"`, otherwise bodies are inspected as one opaque string.

If a CDN or proxy such as Cloudflare fronts the load balancer, rate limits must use `enforce_on_key = "XFF_IP"`, since every request otherwise arrives from a small set of proxy addresses and would share one budget.

### Deployment Notes

`paragon-waf-backendconfig` is added to the `paragon` namespace even when the flag is off, so the first apply after this merge creates that manifest and annotates the public Services. No policy is attached and no Cloud Armor resource is created.

Cloud Armor rate limits are enforced per backend service and per region, so the effective aggregate limit scales with the number of protected backends. Size `waf_rate_limit_global` with that in mind rather than treating it as a single global budget.

### Screenshots

n/a — infrastructure only.

[PARA-21683]: https://useparagon.atlassian.net/browse/PARA-21683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes public ingress security posture when enabled; misconfigured rules or rate limits could block legitimate traffic, though defaults leave WAF off and support preview mode.
> 
> **Overview**
> Adds optional **Google Cloud Armor** WAF and rate limiting for GCP public ingress, mirroring the AWS paragon WAF pattern. Everything stays behind **`waf_enabled`** (default `false`); internal ingress ignores it.
> 
> A new **`gcp/workspaces/paragon/waf`** submodule builds a `google_compute_security_policy` from tfvars: IP allow/deny lists, global and per-path rate limits, Google-managed WAF rule sets, custom CEL rules, and advanced options (JSON parsing, adaptive protection, etc.), with plan-time validation and fixed priority bands.
> 
> The **helm** module always creates **`paragon-waf-backendconfig`** (empty policy name detaches Cloud Armor on disable) and annotates public microservices, monitors (Grafana merges WAF with its health-check BackendConfig), and managed-sync **`api-sync`** so traffic on the shared GCE load balancer is not bypassed. Root **outputs** expose policy name and rule count.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0912577ef874cec5d97ac68c851068487e0de927. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gcp/workspaces/paragon/waf/locals.tf
```diff
@@ -0,0 +1,167 @@
+locals {
+  # Rules are evaluated from the lowest priority number up. Fixed bands leave room to
+  # grow without renumbering neighbouring rules.
```
richard-paragon, line 3: nit: UK/Commonwealth spelling

```suggestion
  # grow without renumbering neighboring rules.
```

richard-paragon (approved):

tedparagon (approved):

PR #206, opened 2026-08-06
feat: allow configuring managed sync RDS instance class

franklingarcia-paragon:
### Issues Closed

- N/A

### Brief Summary

configurable managed sync rds size

### Detailed Summary

The managed sync Postgres instance class was hardcoded to `db.t4g.small`, so larger sizes applied out of band could not be pinned in tfvars and Terraform would drift them back down. This adds `rds_managed_sync_instance_class` (default still `db.t4g.small`) and wires it through the AWS infra postgres module so operators can set the size per workspace.

### Changes

- Add `rds_managed_sync_instance_class` root variable (default `db.t4g.small`)
- Pass it into the postgres module and use it for the managed sync instance size

### Unrelated Changes

N/A

### Future Work

- Consider per-instance size overrides for other RDS roles if needed

### Steps to Test

1. From `aws/workspaces/infra`: `terraform init -backend=false && terraform validate`
2. Set `rds_managed_sync_instance_class` in tfvars and confirm plan only changes managed sync `instance_class`
3. Apply and verify the RDS instance reaches the configured class

### QA Notes

N/A — infra IaC only

### Deployment Notes

- Set `rds_managed_sync_instance_class` in infra tfvars when managed sync should not use the default `db.t4g.small`
- Changing instance class causes an RDS modify (possible reboot/brief downtime)

### Screenshots

N/A

### DIFF
--- a/aws/workspaces/infra/modules.tf
+++ b/aws/workspaces/infra/modules.tf
@@ -45,21 +45,22 @@ module "cloudtrail" {
 module "postgres" {
   source = "./postgres"
 
-  workspace                   = local.workspace
-  aws_region                  = var.aws_region
-  rds_instance_class          = var.rds_instance_class
-  rds_gp3_iops                = var.rds_gp3_iops
-  rds_gp3_storage_throughput  = var.rds_gp3_storage_throughput
-  rds_allocated_storage       = var.rds_allocated_storage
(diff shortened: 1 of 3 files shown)


franklingarcia-paragon:
Superseded by branch feat/PARA-25207/rds-managed-sync-instance-class

PR #224, opened 2026-08-20
fix: harden UDS apply and Grafana sqlite startup

tedparagon:
### Issues Closed

- PARA-24453
- PARA-25666
- PARA-25690

### Brief Summary

harden uds apply and grafana sqlite startup

### Detailed Summary

Two upgrade-time failures: the OpenObserve UDS hook timed out downloading a huge inferred schema and rolled back `paragon-logging`, and Grafana 12 crashed on sqlite `database is locked` during overlapping boot writers (migrations, alerting, provisioning).

The UDS hook is now a Node script (`node:22-alpine`, no `apk add curl jq`) that reads stream settings from the list API (`fetchSchema=false`), never GET `/schema`, and applies in two PUTs. OpenObserve silently drops `defined_schema_fields` names that are not already columns (`save_stream_settings` retain), so a combined `fields` + `defined_schema_fields` PUT on a fresh install kept only the Fluent Bit fields (17 of 494, HTTP 200). Phase 1 creates the columns; phase 2 selects them. Timeouts, 5xx, a missing stream, and partial applies warn and exit 0 so an atomic upgrade cannot roll back. Only missing/invalid config, 401/403, or a rejected payload fail the Job. The Job uses `before-hook-creation` only so logs remain after success and are deleted on the next upgrade.

Grafana gets WAL, higher sqlite retries, more memory, and a Helm fail if `replicaCount` is greater than 1.

Grafana sqlite recovery is PARA-25666 (SEV-976 follow-up). It does **not** switch Grafana to Postgres (PARA-7663). PARA-25690 is the bash → Node rewrite.

Verified on a live cluster from a 17-field UDS: 494 of 494 selected in 1s (`newly added this run: 477`).

### Changes

- Rewrite the UDS hook from Alpine `curl`/`jq` to Node (`node:22-alpine`) with built-in `fetch`
- Apply UDS in two PUTs: create columns first, then set `defined_schema_fields` (OpenObserve otherwise keeps only existing columns)
- Use the streams list API (`fetchSchema=false`) for idempotency; never GET `/schema`
- Retry transient HTTP within a time budget; fail immediately on 401/403 and invalid payloads
- Exit 0 on timeouts, 5xx, missing stream, and partial apply so Helm does not roll back `paragon-logging`
- Keep the hook Job after success (`before-hook-creation` only) so logs can be inspected
- Enable Grafana sqlite WAL and raise query/transaction retries to 50
- Raise Grafana memory to 1Gi / 512Mi
- Fail Helm if Grafana `replicaCount` is greater than 1

### Unrelated Changes

N/A

### Future Work

N/A

### Steps to Test

1. Upgrade `paragon-logging` against a stream that already has a partial UDS (fresh Fluent Bit fields). Confirm logs show `creating 494 column(s)`, then `494 field(s) selected of 494 desired`, then `user-defined schema applied successfully`.
2. Re-upgrade. Confirm `user-defined schema already matches the desired list; nothing to apply`.
3. Confirm Job logs never include a full `/schema` download.
4. Confirm the Job remains after success (`kubectl logs -n paragon job/openobserve-uds-apply`) and is replaced on the next upgrade (`before-hook-creation`).
5. With wrong credentials, confirm the Job fails fast on HTTP 401/403 without printing the password.
6. If OpenObserve is still settling or the stream does not exist yet, confirm the Job warns and exits 0 (Helm release is not rolled back).
7. Upgrade `paragon-monitoring` and confirm Grafana starts (one crash on first boot after a 12.3 annotation backfill is acceptable; it should not stay in CrashLoopBackOff).
8. Confirm the Grafana pod has `GF_DATABASE_WAL=true` and retry env vars, and that `kubectl logs` shows those as config overrides.

### QA Notes

These are Helm hook and monitoring-stack changes, not product UI. Verify from `openobserve-uds-apply` Job logs, Grafana pod logs, and that both releases stay deployed after upgrade.

Live check (partial UDS → full desired list):

```
[openobserve-uds] openobserve endpoint: http://openobserve:5080
[openobserve-uds] stream: default/paragon
[openobserve-uds] desired user-defined schema: 494 field(s) from /uds/openobserve-uds-schema.json
[openobserve-uds] waiting for http://openobserve:5080/healthz (up to 120s)
[openobserve-uds] openobserve is healthy
[openobserve-uds] stream "paragon" currently has 17 user-defined schema field(s)
[openobserve-uds] apply attempt 1/3
[openobserve-uds] step 1/2: creating 494 column(s) on stream "paragon"
[openobserve-uds] step 2/2: selecting 494 field(s) as the user-defined schema
[openobserve-uds] stream "paragon" user-defined schema: 494 field(s) selected of 494 desired
[openobserve-uds]   newly added this run: 477
[openobserve-uds]   already selected before this run: 17
[openobserve-uds] user-defined schema applied successfully
[openobserve-uds] done in 1s
```

### Deployment Notes

No customer values changes required. Clusters stuck on a partial UDS (e.g. 17 of 494) converge on the next `paragon-logging` upgrade.

OpenObserve defaults: `openobserve.uds.image` is `node:22-alpine`, `openobserve.uds.resources.limits.memory=1Gi`, `openobserve.uds.activeDeadlineSeconds=1800`, `openobserve.uds.requestTimeoutSeconds=60`. Private-registry customers must mirror `node` (they previously mirrored `alpine`).

Grafana defaults: `GF_DATABASE_WAL`, `GF_DATABASE_QUERY_RETRIES=50`, `GF_DATABASE_TRANSACTION_RETRIES=50`, memory 1Gi/512Mi.

### Screenshots

N/A

---

> [!NOTE]
> **Medium Risk**
> Changes Helm hook behavior and logging schema apply during upgrades; misconfiguration still fails loudly, but transient OpenObserve issues no longer block the release. Grafana and Terraform fixes are low blast radius.
> 
> **Overview**
> Addresses upgrade-time rollbacks on **`paragon-logging`** and Grafana **`database is locked`** crashes during boot.
> 
> The OpenObserve UDS post-upgrade hook is **replaced** with a Node script on **`node:22-alpine`** (no `curl`/`jq`). It checks idempotency via the streams **list API** with **`fetchSchema=false`**—never **`GET /schema`**—and applies schema in **two PUTs**: add columns, then set **`defined_schema_fields`**. **Config/auth/invalid payloads** fail the Job; **timeouts, 5xx, missing stream, and partial applies** warn and **exit 0** so atomic Helm upgrades do not roll back the release. The hook Job is **retained after success** (`hook-delete-policy: before-hook-creation` only) for **`kubectl logs`**.
> 
> **Grafana** gets SQLite **WAL**, higher query/transaction retries, more memory, and a Helm **fail** if **`replicaCount > 1`**.
> 
> Azure Terraform **ExternalSecret** resources now use **input variables for `count`** instead of `yamlencode(...) != null`, fixing first-apply plan failures when namespace IDs are unknown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 855ab18182184a757f198a0e8d777892b3abee5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -38,63 +67,133 @@ if [ ! -f "$DESIRED_SCHEMA_FILE" ]; then
...
+      return 1
+    fi
+    if extract_paragon "$_dest"; then
+      return 0
+    fi
+    log "stream paragon not listed yet; waiting (${_elapsed}s/${STREAM_WAIT_SECONDS}s)"
+    sleep 5
+    _elapsed=$((_elapsed + 5))
+  done
+  return 2
```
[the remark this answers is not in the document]
tedparagon: Agreed on the split: rollback on legit apply errors, skip only after retries for timeouts / unavailable OpenObserve.

PUT now `warn_skip`s on 404 and the same transient codes as the list path (`000`, `408`, `429`, `5xx`). Other 4xx (`400`, `409`, `422`, …) still `die` so a bad payload fails the hook.

franklingarcia-paragon (approved):

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
```diff
@@ -122,38 +230,42 @@ payload="$(echo "$current" | jq -c --slurpfile d "$DESIRED_SCHEMA_FILE" '
...
   }
-')"
+' "$current" >"$payload"
 
-add_names="$(echo "$payload" | jq '.defined_schema_fields.add | length')"
-remove_names="$(echo "$payload" | jq '.defined_schema_fields.remove | length')"
-remove_fields="$(echo "$payload" | jq '.fields.remove | length')"
+add_names="$(jq '.defined_schema_fields.add | length' "$payload")"
+remove_names="$(jq '.defined_schema_fields.remove | length' "$payload")"
+remove_fields="$(jq '.fields.remove | length' "$payload")"
```
richard-paragon: The variable "$payload" suggests that `payload` is the response body. So passing it to `jq` on the commandline "looks" wrong, because `jq` does not accept JSON content as an argument.

I suggest renaming it to payloadFile to avoid confusion.

richard-paragon (approved):
I'm reluctant about this change, because I think the complexity of applying the UDS now exceeds what is easily maintainable using just curl+jq in a bash script.  

Originally application of the UDS was done using a nodejs script. When Franklin originally re-implemented it as a shell script, I found it slightly hard to reason about (mostly due to the `jq` formulas, which take some thought to interpret).  But at least the bash pathway was fairly clear. With the addition of more complex logic around the requests (with retries, timeouts, and fallback operations), I feel like it's harder to verify the overall logic and core actions.

I would _recommend_ implementing this whole process as a clean nodejs script, where retries/timeouts/fallbacks etc are often a bit easier to express using promises and try/catch etc.

However, I understand that, at this point, doing this might require even more testing and might create other edge cases.

So I am okay going forward with this implementation as long as we can test and verify that it works in the cases listed below:
1. New installation
2. Installation with existing inferred schema (but no UDS)
3. Installation with existing UDS that has fields removed in a subsequent UDS:
   a. UDS (v1): `[ "id", "customerid", "oldid" ]`
   b. UDS (v2): `[ "id", "customerid" ]`
4. Installation with existing UDS with fields that changed type:
   a. UDS (v1): `[ { name: "id" type: "Utf8" }, { name: "correlationid" type: "Int64" }]`
   b. UDS (v2): `[ { name: "id" type: "Utf8" }, { name: "correlationid type: "Utf8" }]`

Case 4 was raised recently, and I *think* it is supported by the existing logic, but we should confirm that.

I think all of these cases can be tested in a single installation. It might only be necessary to destroy/reset the openobserve pod+pvc between tests 1 and 2. Tests 2 - 4 can all be made subsequently without a reset, just by updating the (mock) desired schema.

charts/paragon-logging/charts/openobserve/files/apply-uds.sh
tedparagon: I agree about the complexity of the bash script. I've created a separate ticket for the bash -> node refactor. https://useparagon.atlassian.net/browse/PARA-25690

tedparagon:
Live verification on a cluster that still had the 17-field fresh-install UDS. Two-phase apply selected all 494 desired fields:

```
[openobserve-uds] stream "paragon" currently has 17 user-defined schema field(s)
[openobserve-uds] apply attempt 1/3
[openobserve-uds] step 1/2: creating 494 column(s) on stream "paragon"
[openobserve-uds] step 2/2: selecting 494 field(s) as the user-defined schema
[openobserve-uds] stream "paragon" user-defined schema: 494 field(s) selected of 494 desired
[openobserve-uds]   newly added this run: 477
[openobserve-uds]   already selected before this run: 17
[openobserve-uds] user-defined schema applied successfully
[openobserve-uds] done in 1s
```

This addresses the bash complexity concern (now Node) and the silent drop: OpenObserve only keeps `defined_schema_fields` that already exist as columns, so the apply is two PUTs.

franklingarcia-paragon (approved):

richard-paragon (approved):

PR #233, opened 2026-09-04
PARA-14563: add redis-streams-exporter monitoring chart

franklingarcia-paragon:
### Issues Closed

- [PARA-14563](https://useparagon.atlassian.net/browse/PARA-14563)

### Brief Summary

add redis streams exporter monitoring chart

### Detailed Summary

`redis-streams-exporter` runs on Cloud ECS but has never shipped to Enterprise, so the Grafana Event Logs dashboard renders empty on Kubernetes installs. The monitor was listed in `ignoredServices` in `scripts/update-charts.mjs`, which meant `helm-values.json` was generated without it and no chart existed to deploy.

This adds the chart as a subchart of `paragon-monitoring`, copying the `redis-exporter` and `bull` pattern exactly: the same `paragon-templates` includes, the same `.helmignore`, and one-line template files that delegate to the shared templates. No `ServiceMonitor` is added, because these monitors are scraped by the static jobs Prometheus renders from `MONITOR_<NAME>_HOST` and `MONITOR_<NAME>_PORT`, not by the Prometheus operator.

The earlier attempt at this failed on variable names. The chart declared `MONITOR_REDIS_STREAM_EXPORTER_*` (singular "STREAM") while the exporter and Atlas both use `MONITOR_REDIS_STREAMS_EXPORTER_*`. The env block in `values.yaml` here is generated from `RedisStreamsExporterEnvironmentKeys` in the monorepo, so the names match by construction. The companion monorepo PR closes the loop by registering the monitor with Atlas so this drift is caught by `atlas-env:validate` from now on.

The exporter is enabled by default, matching `redis-exporter`, and is guarded by `subchart.redis-streams-exporter.enabled` so an operator can turn it off without templating errors. Existing installs that do not set the key inherit the default and pick the exporter up on upgrade.

### Changes

- Add `charts/paragon-monitoring/charts/redis-streams-exporter` with `Chart.yaml`, `values.yaml`, `.helmignore`, `NOTES.txt`, `_helpers.tpl`, and one-line deployment, service, serviceaccount, secret, ingress, hpa, and test templates
- Register the subchart in `charts/paragon-monitoring/Chart.yaml` and enable it in `charts/paragon-monitoring/values.yaml`
- Drop `redis-streams-exporter` from `ignoredServices` in `scripts/update-charts.mjs` so the monitor participates in generated helm values
- Add `MONITOR_REDIS_STREAMS_EXPORTER_HOST` and `MONITOR_REDIS_STREAMS_EXPORTER_PORT` (9124) to the AWS, Azure, and GCP `paragon` workspace variables, and to `charts/example.yaml` and `charts/bootstrap/values.yaml`

### Unrelated Changes

None.

### Future Work

Once every monitor is Atlas-registered, `ignoredServices` in `scripts/update-charts.mjs` could go away entirely rather than being pruned one entry at a time.

### Steps to Test

```bash
helm dependency update charts/paragon-monitoring
helm template paragon-monitoring charts/paragon-monitoring -f charts/example.yaml \
  | grep -A5 'redis-streams-exporter'

helm template paragon-monitoring charts/paragon-monitoring -f charts/example.yaml \
  --set 'subchart.redis-streams-exporter.enabled=false' \
  | grep -c 'redis-streams-exporter'
```

What to check:

- the rendered Deployment, Service, ServiceAccount, and Secret carry the `redis-streams-exporter` name and target port 9124
- the Secret exposes `MONITOR_REDIS_STREAMS_EXPORTER_PORT`, `WORKFLOW_REDIS_URL`, `WORKFLOW_REDIS_CLUSTER_ENABLED`, and `WORKFLOW_REDIS_TLS_ENABLED` (plural "STREAMS" everywhere)
- disabling the subchart renders nothing and does not error
- on a real cluster, `helm upgrade` from the previous release adds only the new objects, and Prometheus lists `redis_streams_exporter` as UP with the Event Logs dashboard populated

### QA Notes

This is additive. No existing chart, template, or value is modified beyond registering the new subchart, so an upgrade on an existing Enterprise install should show only the new Deployment, Service, ServiceAccount, and Secret in the diff. Since the subchart defaults to enabled, please confirm the upgrade path on a release that predates this change and does not set the key.

### Deployment Notes

The image comes from the monorepo build and needs the companion monorepo PR merged and released first, since the Kubernetes Prometheus scrape job lives there. Two new environment variables are introduced, `MONITOR_REDIS_STREAMS_EXPORTER_HOST` and `MONITOR_REDIS_STREAMS_EXPORTER_PORT`; both have defaults wired in the three cloud workspaces and in `charts/bootstrap/values.yaml`, so no operator action is required.

### Screenshots

Rendered service from `helm template`:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: redis-streams-exporter
spec:
  ports:
    - port: 9124
      targetPort: 9124
      name: http
```

[PARA-14563]: https://useparagon.atlassian.net/browse/PARA-14563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

image
image


charts/paragon-monitoring/charts/redis-streams-exporter/values.yaml
```diff
@@ -0,0 +1,91 @@
...
+affinity:
+  nodeAffinity:
+    preferredDuringSchedulingIgnoredDuringExecution:
+    - weight: 1
+      preference:
+        matchExpressions:
+        - key: useparagon.com/capacityType
+          operator: In
+          values:
+          - spot
```
richard-paragon, line 89: qq: does this mean it will be 100% spot? I think that's reasonable, since it is a monitor, as long as that means it will be rescheduled to on-demand if there are no spot nodes.

charts/paragon-monitoring/charts/redis-streams-exporter/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1 @@
+{{ include "test-connection.standard" (dict "root" .) }}
```
richard-paragon, line 1: @cursoragent what is involved in the `test-connection.standard`, and should we expect the redis-streams-exporter to support the required endpoints?

richard-paragon (approved):

charts/paragon-monitoring/charts/redis-streams-exporter/templates/tests/test-connection.yaml
richard-paragon: @paragonbot can you answer the question posted above?

charts/paragon-monitoring/charts/redis-streams-exporter/values.yaml
franklingarcia-paragon: Not 100% spot. This is `preferredDuringSchedulingIgnoredDuringExecution` with weight 1, copied from `redis-exporter` (and the other monitors). The scheduler prefers a node labeled `useparagon.com/capacityType=spot`, but if none is available it still schedules on on-demand. It is not `requiredDuringSchedulingIgnoredDuringExecution`, so the pod is never stuck Pending waiting for spot.

If a spot node later disappears, the replica is rescheduled with the same preference (spot if present, otherwise on-demand).

charts/paragon-monitoring/charts/redis-streams-exporter/templates/tests/test-connection.yaml
franklingarcia-paragon: `test-connection.standard` is a Helm test hook (`helm.sh/hook: test`). It starts a one-shot busybox `wget` to `{{ fullname }}:{{ service.port }}` — for this chart that is `redis-streams-exporter:9124`, i.e. `GET /`. It is not run on `terraform apply` / `helm upgrade`; only on an explicit `helm test`.

The exporter serves `/healthz` (`{"ok":true}`) and `/metrics`. There is no `/` handler, so `GET /` is Express 404. `redis-exporter` (oliver006) does answer `/`, which is why the shared test works there.

If we start running `helm test` against monitoring, we should either add a `/` (or `/healthz`) wget override for this chart, or have the exporter treat `/` as health. Not needed for the scrape path we validated on staging (`/metrics` → Grafana Event Logs).

franklingarcia-paragon: Answered on the parent comment: shared test is `wget fullname:9124` → `GET /`. This exporter only has `/healthz` and `/metrics`, so an explicit `helm test` would likely 404. Terraform/Helm upgrade do not run that hook; Prometheus scrapes `/metrics`, which we confirmed on staging.

charts/paragon-monitoring/values.yaml
```diff
@@ -17,6 +17,8 @@ subchart:
     enabled: true
   redis-exporter:
     enabled: true
+  redis-streams-exporter:
```
tedparagon, line 20: nit: this should go after `redis-insight` alphabetically. This would apply throughout.

tedparagon (approved):

PR #234, opened 2026-09-10
feat(aws): add Agent OS infrastructure and External Secrets

franklingarcia-paragon:
### Issues Closed

- PARA-25772
- PARA-25776
- PARA-26290

### Brief Summary

add default-off Agent OS AWS infra and secrets

### Detailed Summary

Adds optional Agent OS data plane on AWS (Postgres, ElastiCache, MSK topics, S3, KMS), dedicated index/extraction capacity, and app/admin secret sync into the `agent-os` namespace with External Secrets plus EKS Pod Identity. Existing Paragon and Managed Sync resources stay on their current addresses. The Helm chart is not installed here.

### Changes

- Dedicated Agent OS RDS (context/tools DBs), ElastiCache, S3 bucket, KMS CMK, and MSK topics, gated by `agent_os_enabled`
- App/admin/vendor payloads in Secrets Manager; ESO merges vendor then app so Terraform keys win collisions, and keeps admin off service pods
- EKS Pod Identity for the Agent OS service account, scoped to the `agent-os` namespace
- Dedicated on-demand `agent-os-index` and `agent-os-extract` managed node groups with workload labels and `NoSchedule` taints
- Matching Karpenter NodePools; extraction defaults to production sizing (`c6a.4xlarge`, maximum 8 nodes)

### Unrelated Changes

N/A

### Future Work

Helm release for Agent OS (PARA-25775)

### Steps to Test

1. `terraform init -backend=false && terraform validate` in `aws/workspaces/infra` and `aws/workspaces/paragon`
2. Plan both workspaces with `agent_os_enabled = false` and confirm no Agent OS resources
3. Plan with `agent_os_enabled = true` and confirm RDS, ElastiCache, S3, secrets, ESO, Pod Identity, and both dedicated node pools appear
4. After apply, confirm Kubernetes secrets `agent-os-app` and `agent-os-admin` in namespace `agent-os`

### QA Notes

Default-off. Infra must apply before paragon. Production extraction capacity adds up to 128 on-demand vCPUs; check quota before enabling. Karpenter can scale its pool to zero, while the managed node group keeps one node. Local `terraform plan` needs valid AWS credentials.

### Deployment Notes

New optional extraction sizing tfvars. Production defaults: `c6a.4xlarge`, min/max 1/8 for the managed node group and max 8 for Karpenter. Staging should override to `c6a.2xlarge` and max 3. No Helm release in this PR. Apply **infra → paragon**.

### Screenshots

N/A

---

> [!NOTE]
> **High Risk**
> Large optional footprint (RDS, cache, MSK identities, dedicated nodes up to production extraction defaults) plus new secrets/KMS/IAM paths; disabling after apply is documented as destructive and infra must apply before paragon.
> 
> **Overview**
> Adds **optional, default-off Agent OS** AWS infrastructure and paragon wiring behind `agent_os_enabled` (requires `managed_sync_enabled`). When enabled, infra provisions a dedicated data plane—KMS CMK, map-driven Postgres (`context`/`tools`), Valkey, versioned S3 with Pod Identity, plus **separate MSK SCRAM users** for app and ACL admin—and seeds **app / admin / vendor** Secrets Manager payloads from module outputs.
> 
> Compute is split into **index** (memory-optimized) and **extract** (compute-optimized) pools with workload labels and `NoSchedule` taints. Infra owns sizing: **managed node groups** when Karpenter is off, or **`agent_os_node_pools` in the cluster/Karpenter handoff** when Karpenter is on (paragon merges those into NodePools and fails apply if they are missing).
> 
> The **paragon** workspace reads the new runtime `agent-os` handoff, overlays paragon-owned secret keys, extends ESO (vendor + app merged into `agent-os-app`, admin separate), grants ESO **KMS decrypt** on the Agent OS key, and associates **EKS Pod Identity** for the `agent-os` service account. **No Agent OS Helm release** in this change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 208f87aa07d719f1fadac1d406ab9bd3a519e776. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aws/workspaces/infra/cluster/locals.tf
```diff
@@ -147,18 +147,72 @@ locals {
     taints          = [local.karpenter_controller_taint]
   }
 
+  # Agent OS index workloads require dedicated on-demand, memory-optimized capacity.
+  agent_os_index_node_group = {
+    min_count      = var.agent_os_index_min_count
+    max_count      = var.agent_os_index_max_count
+    instance_types = var.agent_os_index_instance_types
+    capacity       = "ON_DEMAND"
```
tedparagon, line 155: Why is on-demand required for either of these?

aws/workspaces/infra/cluster/variables.tf
```diff
@@ -147,6 +147,46 @@ variable "use_latest_ami_release_version" {
   default     = false
 }
 
+variable "agent_os_enabled" {
+  description = "Whether to create the Agent OS index managed node group."
+  type        = bool
+  default     = false
```
tedparagon: I would only define defaults on the top level `variables.tf`. These won't be used and introduce added maintenance and confusion.

aws/workspaces/infra/kafka/msk.tf
```diff
@@ -35,9 +35,88 @@ resource "aws_secretsmanager_secret_version" "msk_credentials" {
...
+  cluster_arn = aws_msk_cluster.kafka.arn
+  secret_arn_list = concat(
+    [aws_secretsmanager_secret.msk_credentials.arn],
+    var.agent_os_enabled ? [
+      aws_secretsmanager_secret.msk_credentials_agent_os[0].arn,
+      aws_secretsmanager_secret.msk_credentials_acl_admin[0].arn,
+    ] : []
+  )
+
+  depends_on = [
```
tedparagon: These explicit `depends_on` shouldn't be needed since they are already defined above.

aws/workspaces/infra/kafka/variables.tf
```diff
@@ -36,3 +36,9 @@ variable "msk_autoscaling_enabled" {
   description = "Whether to enable autoscaling for the MSK cluster."
   type        = bool
 }
+
+variable "agent_os_enabled" {
+  description = "Whether to create a dedicated Agent OS SCRAM identity on this MSK cluster."
+  type        = bool
+  default     = false
```
tedparagon: Same comment as above about consolidating defaults in one top level place.

aws/workspaces/infra/postgres/rds.tf
```diff
@@ -151,3 +151,217 @@ resource "aws_db_instance" "postgres" {
...
+  password   = random_password.agent_os_root_password[0].result
+
+  engine               = "postgres"
+  engine_version       = local.agent_os_postgres_config.engine_version
+  instance_class       = local.agent_os_postgres_config.instance_class
+  parameter_group_name = aws_db_parameter_group.agent_os[0].name
+  storage_type         = local.agent_os_postgres_config.storage_type
+
+  iops               = local.agent_os_gp3_striped ? 12000 : null
+  storage_throughput = local.agent_os_gp3_striped ? 500 : null
```
tedparagon: These should probably be variables also since they are regularly tweaked since they require no downtime to change.

aws/workspaces/infra/postgres/rds.tf
```diff
@@ -151,3 +151,217 @@ resource "aws_db_instance" "postgres" {
...
+
+# Agent OS Postgres. Dedicated instance so the Paragon/Managed Sync instance map above
+# keeps its resource addresses; hosts the `context` and `tools` logical databases.
+
+locals {
+  agent_os_postgres_config_defaults = {
+    instance_class         = "db.t4g.medium"
+    allocated_storage      = 100
+    max_allocated_storage  = 1000
+    engine_version         = "16.13"
```
tedparagon: I would just make this "16" since `auto_minor_version_upgrade` is allowed below. Otherwise it will drift and fail to downgrade after a minor upgrade.

aws/workspaces/infra/postgres/rds.tf
```diff
@@ -151,3 +151,217 @@ resource "aws_db_instance" "postgres" {
...
+  deletion_protection    = !var.disable_deletion_protection
+  skip_final_snapshot    = true
+  publicly_accessible    = false
+  storage_encrypted      = true
+  vpc_security_group_ids = [aws_security_group.agent_os[0].id]
+
+  performance_insights_enabled          = true
+  performance_insights_kms_key_id       = var.agent_os_kms_key_arn
+  performance_insights_retention_period = 31
+  enabled_cloudwatch_logs_exports       = ["postgresql", "upgrade"]
```
tedparagon: We probably don't need performance insights on the replica.

aws/workspaces/infra/redis/elasticache.tf
```diff
@@ -201,3 +201,154 @@ resource "aws_appautoscaling_policy" "cache_cpu" {
...
+  replication_group_id = "${substr(replace("${var.workspace}-agent-os", "_", "-"), 0, 32)}-vk"
+  description          = "Agent OS Valkey cache."
+  apply_immediately    = true
+  node_type            = local.agent_os_valkey_config.node_type
+  engine               = "valkey"
+  engine_version       = local.agent_os_valkey_version
+  port                 = 6379
+  parameter_group_name = aws_elasticache_parameter_group.agent_os[0].name
+
+  snapshot_retention_limit = 5
```
tedparagon: This should be at least 7 and maybe 14?

aws/workspaces/infra/redis/elasticache.tf
```diff
@@ -201,3 +201,154 @@ resource "aws_appautoscaling_policy" "cache_cpu" {
...
+    Name    = "${var.workspace}-agent-os-valkey"
+    Cluster = local.agent_os_valkey_config.cluster_enabled ? "true" : "false"
+  }
+}
+
+resource "aws_cloudwatch_log_group" "agent_os" {
+  count = var.agent_os_enabled ? 1 : 0
+
+  name              = "/aws/elasticache/${var.workspace}-agent-os"
+  retention_in_days = 365
```
tedparagon: If these are just system logs above then I'd make it 30 days. Application logs get 365.

aws/workspaces/infra/storage/s3-app.tf
```diff
@@ -212,3 +212,195 @@ resource "aws_iam_role_policy" "app" {
     }
   )
 }
+
+# Agent OS bucket and dedicated Pod Identity role for parsed documents and index data.
+# Static credentials stay out of Agent OS secrets.
+
+resource "aws_s3_bucket" "agent_os" {
```
tedparagon, line 219: This should get `aws_s3_bucket_lifecycle_configuration` also.  Like [here](https://github.com/useparagon/enterprise-installer/blob/main/aws/workspaces/infra/storage/s3-app.tf#L50).

aws/workspaces/infra/app_env.tf
```diff
@@ -114,3 +114,77 @@ locals {
...
+    ADMIN_KAFKA_SASL_MECHANISM = local.agent_os_kafka.acl_admin_kafka_credentials.mechanism
+    ADMIN_KAFKA_SSL_ENABLED    = tostring(local.agent_os_kafka.cluster_tls_enabled)
+
+    KAFKA_PRINCIPAL_ACL_ADMIN    = local.agent_os_kafka.acl_admin_kafka_credentials.username
+    KAFKA_PRINCIPAL_AGENT_OS     = local.agent_os_kafka.agent_os_kafka_credentials.username
+    KAFKA_PRINCIPAL_MANAGED_SYNC = local.agent_os_kafka.kafka_credentials.username
+
+    KAFKA_TOPIC_PARTITIONS          = "3"
+    KAFKA_TOPIC_REPLICATION_FACTOR  = tostring(ceil(var.msk_kafka_num_broker_nodes / 2))
+    KAFKA_TOPIC_MIN_INSYNC_REPLICAS = tostring(ceil(var.msk_kafka_num_broker_nodes / 2))
```
tedparagon, line 188: Not really an infra issue but these variable names are all very generic and could lead to collisions with other apps.

aws/workspaces/infra/variables.tf
```diff
@@ -454,6 +454,103 @@ variable "managed_sync_enabled" {
   default     = false
 }
 
+variable "agent_os_enabled" {
```
tedparagon, line 457: Update tfdoc README with new vars.

aws/workspaces/paragon/helm/external_secrets.tf
```diff
@@ -6,6 +6,23 @@ resource "kubernetes_namespace" "external_secrets" {
   }
 }
 
+resource "kubernetes_namespace" "agent_os" {
+  count = var.agent_os_enabled ? 1 : 0
+
+  metadata {
+    name = "agent-os"
```
tedparagon: I'm torn on having its own namespace. It makes sense but is inconsistent with managed sync. It will also require support training.

aws/workspaces/paragon/variables.tf
```diff
@@ -449,6 +449,57 @@ variable "managed_sync_version" {
...
+variable "agent_os_extract_max_count" {
+  description = "Maximum nodes in the Agent OS Karpenter extraction pool. Use 3 for staging and 8 for production; Karpenter scales to zero when idle."
+  type        = number
+  default     = 8
+
+  validation {
+    condition     = var.agent_os_extract_max_count >= 1
+    error_message = "agent_os_extract_max_count must be at least 1."
+  }
+}
```
tedparagon: I don't love how this is all duplicated from the `infra`. The Karpenter stuff belongs in `infra` but I realize that predates this change.

tedparagon (changes requested):
Various changes requested.

aws/workspaces/infra/cluster/locals.tf
franklingarcia-paragon: This was intentional based on the Agent OS capacity requirement I got from Andrew. These pods are expected to stay up continuously, so I kept these pools on-demand rather than allowing Spot interruptions. @andrew-hernandez-paragon can you confirm that this is still the requirement?

aws/workspaces/infra/storage/s3-app.tf
franklingarcia-paragon: I left the Agent OS bucket lifecycle unchanged for now because I don't want to invent the retention policy. @andrew-hernandez-paragon do we have a minimum retention/lifecycle requirement for the Agent OS S3 data? Once we confirm that, I'll match the rule to the actual data lifecycle instead of copying the app bucket policy blindly.

aws/workspaces/paragon/helm/external_secrets.tf
franklingarcia-paragon: Agreed. I changed this to use the existing Paragon namespace, same pattern as Managed Sync. We don't need a dedicated Agent OS namespace here, and this keeps the support model consistent.

aws/workspaces/paragon/variables.tf
franklingarcia-paragon: Yeah, I agree this duplication isn't ideal. I left the existing Karpenter ownership/layout alone in this PR and only added the Agent OS scheduling/capacity pieces we need. Moving Karpenter fully into infra should be a separate refactor so we don't expand the scope here.

aws/workspaces/infra/app_env.tf
franklingarcia-paragon: Agreed on the naming risk. I left these as-is here because they're the env names expected by Agent OS today, so changing them in infra alone would break the chart/app contract. I think we should namespace them in the app contract first, then update the three cloud installers together.

aws/workspaces/infra/modules.tf
```diff
@@ -190,7 +207,106 @@ module "secrets" {
...
+      type        = "Service"
+      identifiers = ["logs.${var.aws_region}.amazonaws.com"]
+    }
+
+    condition {
+      test     = "ArnEquals"
+      variable = "kms:EncryptionContext:aws:logs:arn"
+      values = [
+        "arn:${data.aws_partition.current.partition}:logs:${var.aws_region}:${data.aws_caller_identity.current.account_id}:log-group:/aws/elasticache/${local.workspace}-agent-os"
+      ]
```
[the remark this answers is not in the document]
franklingarcia-paragon: Good catch. The Valkey config is now a real per-instance map, so additional cache entries can create additional log groups. I updated the KMS encryption-context condition to allow the Agent OS log-group prefix instead of only the original single cache log group.

aws/workspaces/infra/postgres/rds.tf
```diff
@@ -151,3 +151,205 @@ resource "aws_db_instance" "postgres" {
...
+  vpc_security_group_ids = [aws_security_group.agent_os[0].id]
+
+  enabled_cloudwatch_logs_exports = ["postgresql", "upgrade"]
+
+  apply_immediately = true
+
+  tags = {
+    Name = "${local.agent_os_postgres_names[each.key]}-replica"
+  }
+}
```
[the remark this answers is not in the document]
franklingarcia-paragon: Fixed. The replica now references the primary DB identifier and only sets replica-specific attributes so RDS can inherit engine/storage settings from the source.

aws/workspaces/infra/redis/elasticache.tf
```diff
@@ -201,3 +201,149 @@ resource "aws_appautoscaling_policy" "cache_cpu" {
...
+  tags = {
+    Name = "${local.agent_os_valkey_names[each.key]}-valkey"
+  }
+}
+
+resource "aws_elasticache_replication_group" "agent_os" {
+  for_each = local.agent_os_valkey_instances
+
+  # Replication group IDs are capped at 40 characters.
+  replication_group_id = "${substr(replace(local.agent_os_valkey_names[each.key], "_", "-"), 0, 32)}-vk"
```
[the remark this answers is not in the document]
franklingarcia-paragon: Fixed. I split the short ElastiCache resource ID from the full logical name and reserve space for a per-key hash on additional map entries, so long workspace prefixes can't collapse multiple Valkey instances to the same ID.

aws/workspaces/infra/storage/s3-app.tf
franklingarcia-paragon: Fixed. I added lifecycle hygiene for the Agent OS bucket: incomplete multipart uploads are aborted after 1 day, noncurrent versions expire after 30 days, and expired delete markers are removed. I intentionally left current-object expiration out so infrastructure does not invent an application retention policy.

aws/workspaces/infra/storage/s3-app.tf
```diff
@@ -212,3 +212,195 @@ resource "aws_iam_role_policy" "app" {
...
+}
+
+resource "aws_s3_bucket_versioning" "agent_os" {
+  count  = var.agent_os_enabled ? 1 : 0
+  bucket = aws_s3_bucket.agent_os[0].id
+
+  versioning_configuration {
+    status = "Enabled"
+  }
+}
```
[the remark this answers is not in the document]
franklingarcia-paragon: Fixed in 807eae83. The Agent OS bucket now has lifecycle rules for incomplete multipart uploads, noncurrent versions, and expired delete markers while leaving current-object retention to the application contract.

aws/workspaces/paragon/variables.tf
franklingarcia-paragon: Fixed this in a98ae00e. I kept the existing Karpenter resource ownership in the paragon workspace, but moved the Agent OS pool configuration itself into infra. Infra now owns the Agent OS instance types, limits, capacity type, labels, and taints and publishes the resulting node-pool config through the existing cluster handoff. Paragon only renders those NodePools from the infra-provided config, so there is no second Agent OS source of truth.

aws/workspaces/infra/storage/s3-app.tf
andrew-hernandez-paragon: yeah, just no `expiration { days }`.

`parsed/` is the extracted text from customer files for hydrating search.

`index/` is the lance search index, ie the live database. expiring it deletes files that are still in use.

so cleanup rules only: incomplete uploads 1 day, old versions 30 days on `parsed/` and 7 days on `index/`, delete markers on.

aws/workspaces/infra/cluster/locals.tf
andrew-hernandez-paragon: yea i suggested on demand because there is expensive work that would cost us alot of recompute if aws reclaimed so the 2 new node groups should be on demand the shared one used by a majority of the services is spot

aws/workspaces/paragon/helm/external_secrets.tf
```diff
@@ -6,6 +6,15 @@ resource "kubernetes_namespace" "external_secrets" {
...
 }
 
+resource "kubernetes_service_account" "agent_os" {
+  count = var.agent_os_enabled ? 1 : 0
+
+  metadata {
+    name      = "agent-os"
+    namespace = local.paragon_namespace
+  }
+}
```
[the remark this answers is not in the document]
franklingarcia-paragon: Fixed in dab92c31. Agent OS now has its own `agent-os` namespace; the KSA, ExternalSecrets, and secret reads use that namespace, Pod Identity consumes the Agent OS namespace output, and the S3 role trust policy now requires `kubernetes-namespace=agent-os`.

aws/workspaces/infra/app_env.tf
franklingarcia-paragon: @andrew-hernandez-paragon @adrian-paragon any thoughs on this? can we add a prefix like `AOS_` and ensure that the app use it?

PR #240, opened 2026-09-17
PARA-25647: Add on-prem OCS knative code-runner

anand-paragon:
PARA-25647

## Summary
- Add `paragon-onprem` subchart `ocs-code-runner` for Docker Hub `useparagon/ocs-code-runner-knative` so on-prem can run `CODE_RUNTIMES=knative` without ECR/Lambda.
- Default `mode: deployment` is a ClusterIP Service `80→8080` (no Knative CRDs). Optional `mode: knative` emits a `serving.knative.dev/v1` Service for scale-to-zero.
- Workers receive `WORKER_SHARED_OCS_KNATIVE_SERVICE_URL=http://ocs-code-runner.paragon.svc.cluster.local` plus `OCS_CODE_RUNNER_STORAGE_*` (same provider set as `CLOUD_STORAGE_TYPE`). AWS/GCP/Azure list the microservice for pod/workload identity; `update-charts` ignores `ocs-code-runner` (not an atlas service).

## Test plan
- [ ] After substituting `__PARAGON_VERSION__` → `0.0.0`, `helm template` in deployment mode shows image `useparagon/ocs-code-runner-knative:latest` and ClusterIP `80→8080`.
- [ ] Worker-actions env includes `WORKER_SHARED_OCS_KNATIVE_SERVICE_URL` and `OCS_CODE_RUNNER_STORAGE_*` when those keys are in `global.env`.
- [ ] `mode: knative` with `CLOUD_STORAGE_TYPE=GCP` emits `apiVersion: serving.knative.dev/v1` Service named `ocs-code-runner`, `OCS_CODE_RUNNER_STORAGE_TYPE=gcp`, containerPort `8080` / `http1`.
- [ ] Existing clusters without Knative Serving can helm-upgrade (default mode does not require CRDs).
- [ ] Confirm this does **not** change platform-monorepo `createOcsRuntime` (still Lambda-only until a later platform PR).

Made with [Cursor](https://cursor.com)


aws/workspaces/paragon/helm/knative.tf
```diff
@@ -0,0 +1,107 @@
...
+# ocs-code-runner (serving.knative.dev/v1 Service). Kourier is ClusterIP so
+# apply does not provision a public load balancer; workers call the cluster-local URL.
+
+resource "kubernetes_namespace" "knative_operator" {
+  metadata {
+    name = "knative-operator"
+  }
+}
+
+resource "helm_release" "knative_operator" {
```
tedparagon: Will OCS always be deployed or should it have a flag to opt-in like Managed Sync?

aws/workspaces/paragon/variables.tf
```diff
@@ -674,6 +674,11 @@ locals {
       "port"             = try(local.helm_vars.global.env["FLIPT_PORT"], 1722)
       "public_url"       = try(local.helm_vars.global.env["FLIPT_PUBLIC_URL"], null)
     }
+    "ocs-code-runner" = {
+      "healthcheck_path" = "/healthz"
+      "port"             = 8080
```
tedparagon: Should this be a custom port like 17xx?

aws/workspaces/paragon/variables.tf
```diff
@@ -1075,6 +1080,10 @@ locals {
             local.cloud_storage_type == "S3" ? "https://s3.${var.aws_region}.amazonaws.com" : null,
           )
 
+          # OCS knative code-runner (on-prem substitute for WORKER_SHARED_OCS_LAMBDA_NAME).
+          # Storage offload uses CLOUD_STORAGE_* (same contract as platform createOcsRuntime).
+          WORKER_SHARED_OCS_KNATIVE_SERVICE_URL = try(local.helm_vars.global.env["WORKER_SHARED_OCS_KNATIVE_SERVICE_URL"], "http://ocs-code-runner.paragon.svc.cluster.local")
```
tedparagon: We typically don't provide the FQDN and just rely on the local name and port.
```
      WORKER_ACTIONS_PRIVATE_URL:                      http://worker-actions:1712
      WORKER_CREDENTIALS_PRIVATE_URL:                  http://worker-credentials:1713
      WORKER_DEPLOYMENTS_PRIVATE_URL:                  http://worker-deployments:1718
      WORKER_EVENT_LOGS_PRIVATE_URL:                   http://worker-eventlogs:1723
      WORKER_PROXY_PRIVATE_URL:                        http://worker-proxy:1715
```

charts/paragon-templates/templates/_env.yaml
```diff
@@ -32,4 +32,10 @@ Environment variables helper
       key: {{ $key }}
       optional: true
 {{- end }}
+{{- /* OCS knative URL: inject even before atlas-env.yaml lists the key.
+     Storage offload uses CLOUD_STORAGE_* already present on workers. */}}
+{{- if and $.Values.global.env (kindIs "map" $.Values.global.env) (hasKey $.Values.global.env "WORKER_SHARED_OCS_KNATIVE_SERVICE_URL") (not (has "WORKER_SHARED_OCS_KNATIVE_SERVICE_URL" $envKeys)) }}
+- name: WORKER_SHARED_OCS_KNATIVE_SERVICE_URL
+  value: {{ index $.Values.global.env "WORKER_SHARED_OCS_KNATIVE_SERVICE_URL" | quote }}
+{{- end }}
```
tedparagon: I'm not really sure why this is needed but I don't like the idea of having to special case one envKey in our generic template. Why doesn't this one key need to behave differently than all the rest?

charts/paragon-onprem/charts/ocs-code-runner/templates/ksvc.yaml
```diff
@@ -0,0 +1,50 @@
...
+          image: {{ include "paragon.image" (dict "root" . "repository" .Values.image.repository "tag" .Values.image.tag) | quote }}
+          imagePullPolicy: {{ include "paragon.imagePullPolicy" (dict "root" .) }}
+          ports:
+            - containerPort: {{ .Values.service.port }}
+              name: http1
+          env:
+            - name: NODE_OPTIONS
+              value: --max-old-space-size=768
+          readinessProbe:
+            httpGet:
```
franklingarcia-paragon: Both HTTP probes are missing `httpGet.port`. Kubernetes requires a port for an HTTP probe, so can we set this to `http1` or `{{ .Values.service.port }}` for both readiness and liveness?

charts/paragon-onprem/charts/ocs-code-runner/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  type: ClusterIP
+  ports:
+    # Port 80 so WORKER_SHARED_OCS_KNATIVE_SERVICE_URL needs no :8080 suffix.
+    - port: 80
+      targetPort: 8080
+      protocol: TCP
+      name: http
+  selector:
+    serving.knative.dev/service: {{ include "ocs-code-runner.ksvcName" . }}
```
franklingarcia-paragon: I don't think we should route directly to the revision pods here. This Service targets the user container on `8080`, which bypasses Knative's normal request path/queue-proxy and its concurrency, timeout, draining, and revision traffic routing behavior. Can we keep a stable in-cluster endpoint but have it enter through the Knative routing path instead of selecting the revision pods directly?

charts/paragon-onprem/charts/ocs-code-runner/values.yaml
```diff
@@ -0,0 +1,48 @@
+# OCS knative node-script code runner (on-prem substitute for AWS Lambda).
+# Emits a cluster-local serving.knative.dev/v1 Service. Requires Knative Serving + Kourier.
+
+image:
+  # Docker Hub equivalent of ocs/knative-node-script. Independent of Paragon VERSION.
+  repository: useparagon/ocs-code-runner-knative
+  pullPolicy: IfNotPresent
+  tag: latest
```
franklingarcia-paragon: Can we pin this image to a release tag or digest instead of `latest`? This chart calls `paragon.image` directly rather than `paragon.containerImage`, so this really does deploy `:latest`. Combined with `IfNotPresent` and the Docker Hub tag-resolution skip in the Knative config, different nodes can run different runner builds for the same Enterprise release.

aws/workspaces/paragon/helm/knative.tf
```diff
@@ -0,0 +1,107 @@
...
+  depends_on = [
+    helm_release.knative_operator,
+    kubernetes_secret.knative_serving_image_pull,
+  ]
+}
+
+# Operator installs serving.knative.dev CRDs asynchronously after the CR is applied.
+resource "time_sleep" "wait_for_knative_serving_crds" {
+  depends_on      = [kubectl_manifest.knative_serving]
+  create_duration = "90s"
```
franklingarcia-paragon: Can we wait on an actual Knative readiness condition instead of a fixed 90s sleep? Operator reconciliation is asynchronous, so a slow first install can exceed this and Helm will start applying `serving.knative.dev` resources before Serving is ready. The same pattern exists in Azure/GCP; waiting for the CRD or `KnativeServing` Ready condition would make this deterministic.

charts/paragon-onprem/charts/ocs-code-runner/templates/service.yaml
anand-paragon: ✅

anand-paragon: Now request go through proxy, so autoscaling and metrics collection works , i caught this while testing grafana metrics

charts/paragon-onprem/charts/ocs-code-runner/values.yaml
anand-paragon: This has been kept as per consistency from other charts like `hermes` , also the version of `ocs_code_runner` is different then typical 
```
global:
  env:
    VERSION: xxx
```

It will go separately as in `.Values.yaml`
```
ocs-code-runner:
  image:
    tag: v0.9.4-canary.8
```

aws/workspaces/paragon/helm/knative.tf
anand-paragon: It will always be deployed, since it will be runtime for integration intents

charts/paragon-templates/templates/_env.yaml
anand-paragon: Hey @tedparagon  i did not want to update the AMI, it's just for testing , i will remove it today i am almost done with my testing .

aws/workspaces/paragon/variables.tf
anand-paragon: I would say let's reserve 17xx for only paragon known microservices, `ocs code runner` is a `knative` code runner.

aws/workspaces/paragon/variables.tf
anand-paragon: ✅ Done

charts/paragon-onprem/charts/ocs-code-runner/templates/ksvc.yaml
anand-paragon: ✅ Done

aws/workspaces/paragon/helm/knative.tf
anand-paragon: ✅ Done

charts/paragon-templates/templates/_env.yaml
anand-paragon: reverted now

franklingarcia-paragon (approved):

tedparagon (approved):

